PR #12, opened 2025-10-05
[F4KRP-85] Create jobs table

Kerushani:
## JIRA ticket link
[Create jobs table](https://f4kblueprint.atlassian.net/browse/F4KRP-85)

- Added jobs table as outlined in the [FigJam](https://www.figma.com/board/MpiIlmgUQusW0aNiOU6cBo/F4K-Schema-FigJam?node-id=0-1&p=f&t=IenZpiDjnYv4dPVa-0)

## Steps to test
1. Review `jobs.py` and `num.py`

* Check if all columns are appropriately added (can refer back to FigJam)
* Do data types make sense; do default values makes sense

## Checklist
- [x] I have run the appropriate linter(s)


ChloeW125 (approved):

backend/python/app/models/jobs.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class JobsBase(SQLModel):
+    """Shared fields between table and API models"""
+
+    route_group_id: UUID | None = Field(foreign_key="route_groups.id")
+    progress: ProgressEnum = Field(default=ProgressEnum.PENDING)
+
+
+class Jobs(JobsBase, BaseModel, table=True):
```
ColinToft, line 17: I think I would call this class `Job` to avoid confusion, because it will be used in other code (like services) represent just one job (instead of the whole table).

backend/python/app/models/jobs.py
```diff
@@ -0,0 +1,53 @@
...
+from sqlmodel import Field, SQLModel
+
+from .base import BaseModel
+from .enum import ProgressEnum
+
+
+class JobsBase(SQLModel):
+    """Shared fields between table and API models"""
+
+    route_group_id: UUID | None = Field(foreign_key="route_groups.id")
```
ColinToft, line 13: Based on the decision we were talking about at the end of last meeting I think this should be "route_groups.route_group_id"

backend/python/app/models/jobs.py
```diff
@@ -0,0 +1,53 @@
...
+
+class Jobs(JobsBase, BaseModel, table=True):
+    """Jobs model for demonstration purposes"""
+
+    __tablename__ = "jobs"
+
+    job_id: UUID = Field(default_factory=uuid4, primary_key=True)
+    started_at: datetime | None = Field(
+        default_factory=datetime.utcnow,
+    )
```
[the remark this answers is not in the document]
ColinToft: I think we can just default this to None, and then have whichever code handling the Job set the `started_at` field when the job starts! And that will help avoid this utcnow thing (lowkey idk what this is saying)

ColinToft (changes requested):
Looks great! A couple comments from me + Copilot but after that should be good to go!

ColinToft (approved):
LGTM!

PR #55, opened 2025-11-13
Driver History API Routes

ludavidca:
## JIRA ticket link
[API route to update an entry in the driver history table](https://f4kblueprint.atlassian.net/jira/software/projects/F4KRP/boards/1?selectedIssue=F4KRP-45)

## Implementation description
* Added Get, Post, Patch, and Delete Routes to driver history table

## Steps to test
1. Build application in docker
2. Get authenticated, use fastapi docs to add a couple driver history to the table
3. Test API Routes to ensure everything functions as intended

* correct http error codes and error handling
* logic of constructed api routes

## Checklist

### DIFF
--- a/backend/python/app/models/driver_history.py
+++ b/backend/python/app/models/driver_history.py
@@ -17,16 +17,8 @@ class DriverHistoryBase(SQLModel):
     """Shared fields between table and API models"""
 
     driver_id: UUID = Field(foreign_key="drivers.driver_id", index=True)
-    year: int = Field(nullable=False)
+    year: int = Field(nullable=False, ge=MIN_YEAR, le=MAX_YEAR )
     km: float = Field(nullable=False)
-
-    @field_validator("year")
-    @classmethod
-    def validate_year(cls, v: int) -> int:
-        if not (MIN_YEAR <= v <= MAX_YEAR):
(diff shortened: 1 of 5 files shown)


backend/python/app/models/driver_history.py
```diff
@@ -17,17 +17,9 @@ class DriverHistoryBase(SQLModel):
...
     km: float = Field(nullable=False)
 
-    @field_validator("year")
-    @classmethod
-    def validate_year(cls, v: int) -> int:
-        if not (MIN_YEAR <= v <= MAX_YEAR):
-            raise ValueError("Year must be between 2025 and 2100")
-        return v
-
-
```
ColinToft: What is the rationale for removing this?

ColinToft (changes requested):
Good stuff David! One comment from me and a few copilot comments, otherwise LGTM!

backend/python/app/models/driver_history.py
ludavidca: I tried getting it to work, but v kept on being read as 0. I instead elected to do validation during the api call, and I also implemented ge (greater than or equal to) and le (lower than or equal to) on the year field, which is much shorter and concise. This ensures that there are two levels of validations

ColinToft (approved):
LGTM!

PR #66, opened 2025-11-24
Add sweep clustering

Kerushani:
## JIRA ticket link

[F4KRP-130](https://f4kblueprint.atlassian.net/browse/F4KRP-130)

## Implementation description
* This PR adds SweepClusteringAlgorithm, a clustering implementation that groups a list of Location objects into clusters around a warehouse coordinate.
* Computes each location’s polar angle relative to the warehouse using atan2, normalized to [0, 2π) using math.tau.

## Steps to test
There's a test file (sweep_algorithm_test.py) - there are variables on Lines 33 - 35 that can be adjusted - feel free to test it out with that for different cluster numbers, locations per cluster, and max boxes per cluster.

The test can be run with `docker compose exec backend sh -lc 'python -m app.services.implementations.sweep_algorithm_test'`

Correctness of sweep ordering implementation
- Angle computation (atan2(lat_diff, lon_diff))
- Secondary sort by distance squared

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/sweep_algorithm_test.py
+++ b/backend/python/app/services/implementations/sweep_algorithm_test.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+"""
+Test script for KMeans clustering with real database locations.
+Run with: python -m app.services.implementations.k_means_test
+"""
+
+import os
+import sys
+
+import matplotlib.pyplot as plt
+import pandas as pd  # Often useful for data handling
+import seaborn as sns
+
+from app.utilities.geocoding import geocode
+
+sys.path.insert(0, "/app")
+
+from app.models.admin import Admin
+from sqlmodel import Session, create_engine, func, select
+
+# Import all models to register them with SQLModel
+from app.models.location import Location
+from app.models.location_group import LocationGroup  # noqa: F401
+from app.models.route import Route  # noqa: F401
+from app.models.route_group import RouteGroup  # noqa: F401
+from app.models.route_group_membership import RouteGroupMembership  # noqa: F401
+from app.models.route_stop import RouteStop  # noqa: F401
+from app.services.implementations.sweep_clustering import (
+    SweepClusteringAlgorithm,
+)
+
+# Use the same connection string as seed_database.py
+DATABASE_URL = "postgresql://postgres:postgres@f4k_db:5432/f4k"
+
+
+def main() -> None:
+    engine = create_engine(DATABASE_URL, echo=False)
+
+    with Session(engine) as session:
+        # Fetch locations that have coordinates
+        statement = (
+            select(Location)
+            .where(Location.latitude is not None, Location.longitude is not None)
+            .order_by(func.random())
+            .limit(20)
+        )
+
+        locations = list(session.exec(statement).all())
+
+        print(f"Fetched {len(locations)} locations from database\n")
+
+        if len(locations) < 2:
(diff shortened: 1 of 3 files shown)


ludavidca (changes requested):
Overall Implementation seems solid, but there are still some lint issues!  Please feel free to resolve them and request review again!

eddywang4340 (approved):
Algorithm LGTM just not sure if we should keep that test file since I believe there is an initiative to add tests already

ludavidca:
@claude review

backend/python/app/services/implementations/sweep_algorithm_test.py
```diff
@@ -0,0 +1,178 @@
...
+
+async def main() -> None:
+    engine = create_engine(DATABASE_URL, echo=False)
+
+    with Session(engine) as session:
+        # Fetch locations that have coordinates
+        statement = (
+            select(Location)
+            .where(Location.latitude is not None, Location.longitude is not None)
+            .order_by(func.random())
```
ludavidca: Yeah, I think Claude (or Colin) might be right here:

.where(Location.latitude is not None, Location.longitude is not None)
Location.latitude is not None uses Python identity comparison against a SQLModel/SQLAlchemy column descriptor, which always evaluates to True at the Python level. No filtering is actually applied. This should be:

.where(Location.latitude.is_not(None), Location.longitude.is_not(None))

ludavidca (changes requested):
Left Comments and Good Work!

backend/python/app/services/implementations/sweep_clustering.py
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(self, warehouse_lat: float, warehouse_lon: float) -> None:
+        self._warehouse_lat = warehouse_lat
+        self._warehouse_lon = warehouse_lon
+
+    async def cluster_locations(
+        self,
+        locations: list[Location],
+        num_clusters: int,
+        max_locations_per_cluster: int | None = None,
+        max_boxes_per_cluster: int | None = None,
```
ludavidca: Small Food 4 Thought, should we even include max_boxes_per_cluster as a parameter or just retrieve it from system settings? (No need to fix, just want to get an opinion)

default_cap is in system_settings

backend/python/app/services/implementations/sweep_algorithm_test.py
```diff
@@ -0,0 +1,237 @@
...
+            print(f"Unexpected error (PACKED): {e}")
+            import traceback
+
+            traceback.print_exc()
+
+
+if __name__ == "__main__":
+    import asyncio
+
+    asyncio.run(main())
```
ludavidca: Move file to backend/python/scripts after pulling from main

backend/python/app/services/implementations/sweep_clustering.py
Kerushani: Good point - but just wanted to bring up that the `ClusteringAlgorithmProtocol` says `Algorithms should not interact with the database - they only compute the cluster assignments`.

In this case, I was thinking that calling `system_settings` directly in `sweep_clustering.py` would break the protocol. We could eventually use `default_cap` when sweep is called somewhere is the service layer.

Does this checkout?

ludavidca: Yes that makes sense! That is a good idea, let's implement it in the service layer

ludavidca (approved):
LGTM

PR #74, opened 2025-12-30
Add more tests

ColinToft:
## Implementation description
* Adds more tests to API routes
* Add a test for the seeding script
* Fixes issue where tests used the actual DB (not the test DB) which wiped out actual info stored in the local DB

## Steps to test
1. Ensure all API routes are covered

* 

## Checklist
- [ ] My PR name is descriptive and in imperative tense
- [ ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


ludavidca:
@claude please review this PR

backend/python/tests/data/test_locations.csv
ludavidca: Just to confirm, we're cool with leaving this in right?

backend/python/tests/test_routes.py
```diff
@@ -0,0 +1,421 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+
+class TestDriverRoutes:
+    """Test suite for driver API routes."""
+
+    @pytest.mark.asyncio
+    async def test_get_drivers_empty(self, async_client: AsyncClient) -> None:
+        """Test GET /drivers returns empty list when no drivers exist."""
+        response = await async_client.get("/drivers/")
```
ludavidca: I assume that these tests are only to be run when pushed to github actions, so there's no need for them to be run locally. If that's the case, then this should be fine, or else the json is probably not empty lol

ludavidca (changes requested):
Left a couple of comments, but should be fine once we address C Dawg's (Claude's) review + a couple of small nits from me. Happy New Year!

ludavidca (approved):
LFGTM

PR #84, opened 2026-02-08
Finish Up Restructuring User System

kingMonkeh:
## JIRA ticket link
[Finish Up Restructuring User System](https://f4kblueprint.atlassian.net/issues?jql=assignee%20%3D%20currentUser()%20AND%20project%20%3D%20F4KRP%20AND%20statusCategory%20!%3D%203&selectedIssue=F4KRP-140)

## Implementation description
* Wrote small python script to go through all firebase users and set to 'driver' role, except for admin user food4kids@uwblueprint.org
* deleted oauth routes since we decided we are not using them
* reenabled require_authorization_by_role
* added cascade delete to driver, admin

## Steps to test
1. I already ran the python script so you don't need to run it again unless you made a new account without firebase roles yet
2. update your local db user role to be 'driver' and sign in with your account, alternative is to use the seeded admin user from seed_database.py
3. with your auth token try hitting the example admin only route GET http://localhost:8080/admins/test, should return a string if on the admin account, otherwise should return forbidden

* checking that requiring role is working properly
* using delete user as the main source of truth, ie should I get rid of delete_driver service then ? since you should always be deleting the user to cascade

## Checklist


kingMonkeh:
@claude

backend/python/app/update_firebase.py
```diff
@@ -0,0 +1,44 @@
+from firebase_admin import auth
+
+from app import initialize_firebase
+
+admin_email = "food4kids@uwblueprint.org"
+
```
ludavidca: Might be best to have this as an env variable to avoid hardcoded values

backend/python/app/update_firebase.py
```diff
@@ -0,0 +1,44 @@
...
+        page = page.get_next_page()
+
+    print(f"\nSuccessfully updated {count} users.")
+
+
+if __name__ == "__main__":
+    # Change to new desired role
+    # NOTE: This overwrites preexisting roles so be careful!
+    new_role = "driver"
+    update_all_users_role(new_role)
```
ludavidca: Once the update_your_user_role is implemented (with input param being your email). Please feel free to make that the default __main__ function that runs, and make a variable your_email  = "" to be used by the function!

ludavidca (changes requested):
Great Job making these changes! User System is definitely one of the most critical and important and complex parts of the project and I'm really happy with how much work you've put into this. There are still a couple of small comments and nice to haves. Lmk your thoughts!

kingMonkeh:
@claude review

ludavidca (approved):
Great Work Hy Lac, Looks great to me!

PR #85, opened 2026-02-10
GET routes/{route_id}

Artyom-G:
## JIRA ticket link

[Ticket Name](https://f4kblueprint.atlassian.net/jira/software/projects/F4KRP/boards/1?jql=assignee%20%3D%20712020%3A83222178-947d-4799-bcde-9842e4dbda34&selectedIssue=F4KRP-120)

## Implementation description
* Added a GET routes/{route_id} route

## Steps to test
1. Pick an arbitrary route_id (you can select this by testing GET /routes)
2. Test GET /routes/{route_id}
3. Both endpoints require auth (in theory) but this seems to fail, I can work on that but under a different ticket since that seems like an auth issue

* I was following the other routes as a guide, I wonder if this is the best way to do it. Look at my TODO comment under the DELETE routes/{route_id}:

```python
    async def delete_route(self, session: AsyncSession, route_id: UUID) -> bool:
        """Delete route by ID"""
        try:
            statement = select(Route).where(Route.route_id == route_id)
            result = await session.execute(statement)
            route = result.scalars().first()

            if not route:
                self.logger.error(f"Route with id {route_id} not found")
                return False

            await session.delete(route)
            await session.commit()

            return True
        except Exception as error:
            self.logger.error(f"Failed to delete route {route_id}: {error!s}")
            await session.rollback()
            # TODO: do we really want to return the raw error
            raise error
```
* NOTE: This is not my code, I used this as a reference

## Checklist


ludavidca (changes requested):

Artyom-G:
implemented all your changes

ludavidca (changes requested):
Small Change

backend/python/app/routers/route_routes.py
```diff
@@ -38,6 +38,26 @@ async def get_routes(
...
+        route_id (UUID): The unique identifier of the route to delete.
+        session (AsyncSession): The database session dependency.
+
+    Returns:
+        None. Responds with HTTP 200 OK on successful get.
+    """
+
+    route = await route_service.get_route(session, route_id)
+    return route
+
```
ludavidca: 3. HTTPException belongs in the router, not the service (route_service.py)

Following the pattern already established by delete_route (which returns a plain bool), the service layer should raise a domain-level exception or return None. At the same time, the router is responsible for translating that into an HTTP response. Coupling the service to FastAPI's HTTPException makes the service harder to test and reuse.

A minimal fix consistent with the existing delete_route pattern:

# service: return None on not-found, let DB errors propagate
async def get_route(self, session: AsyncSession, route_id: UUID) -> Route | None:
    try:
        statement = select(Route).where(Route.route_id == route_id)
        result = await session.execute(statement)
        return result.scalars().first()
    except Exception as error:
        self.logger.exception("Failed to get route %s", route_id)
        await session.rollback()
        raise

# router: convert to HTTP response
@router.get("/{route_id}", response_model=Route, status_code=status.HTTP_200_OK)
async def get_route(route_id: UUID, session: AsyncSession = Depends(get_session)) -> Route:
    route = await route_service.get_route(session, route_id)
    if not route:
        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Route {route_id} not found")
    return route


This also resolves the 404 error mentioned previously (and my bad on misguiding you, 404 error should be raised here instead)

PR #106, opened 2026-03-14
F4KRP-126 Added edit functionality for routes

eddywang4340:
## JIRA Ticket
[Functionality to edit routes: PATCH /routes/{id}](https://f4kblueprint.atlassian.net/browse/F4KRP-126)

## Implementation Description

- Added a Patch Request class in the Route model (notice that all fields are optional because you can update no fields or you can update all fields)
- Added a patch route for updating a route that calls the service to update a route
- Added route service to update a route
  - If name is provided in patch, update it
  - If notes is provided in patch, update it
  - If location ids is provided in patch, do the following:
    -  Validate all location ids provided exist (if not raise error - aka no update made)
    - If they exist, fetch warehouse coordinates and call fetch route polyline to generate a new polyline for the new location ids
    - Delete existing route stops (note: we can't delete the entire route itself or else all existing fields will be lost, so we iterate through the route stops table)
    - Create new route stops from the given location ids (with stop number from what the user specified)
    - Update the polyline string itself, the updated_at field, and the route length

## Steps to Test

1.

*

## Checklist


ludavidca (approved):
Great work, Eddy! Don't worry about the Notes Comment, as I think we will get to that when we get a finalized design. Feel free to merge when ready!

PR #107, opened 2026-03-17
[F4KRP-116] Auth middleware 

Jam-Cai:
## Implementation Description

* Added auth dependencies in auth.py, role-based (`require_admin`, `require_authorization_by_role`), identity-based (`require_authorization_by_user_id`, `require_authorization_by_email`), and token extraction helpers (`get_current_user_id`,  `get_current_user_email, get_current_database_user_id`).
* Added role based route protection
* Added unittests
## Steps to Test

1. run docker-compose exec backend python -m pytest tests/test_auth_middleware.py -v --no-cov
2. login as a user and test rbac gated endpoints

* double check if roles for endpoints are correct

## Checklist


backend/python/app/routers/database_routes.py
```diff
@@ -26,6 +27,7 @@ class TableDataInfo(TypedDict):
 @router.get("/table")
 async def get_all_tables(
     session: AsyncSession = Depends(get_session),
+    _auth: bool = Depends(require_driver_or_admin),
 ) -> Response:
```
ludavidca: I think this should actually be an admin only endpoint, and we should also remove this method as http access to DB is a pretty big security risk. Might be good to delete the entire file.

backend/python/app/dependencies/auth.py
```diff
@@ -101,6 +103,145 @@ async def check_user_id(
     return check_user_id
 
 
+def require_self_or_admin(user_id: str) -> Callable:
+    """
+    Create a dependency that allows access if the user is an admin,
```
ludavidca: This is defined and should definitely be used.(Currently dead code) This would be really helpful for some of the comments I raised below!

ludavidca (changes requested):
Hey James, thanks for putting this up! I've gone through and left comments throughout.

Honestly, I think a good chunk of the issues come down to me not being as clear enough in the ticket requirements. That's on me, and I want to make sure we get aligned before you spend more time on revisions.

There are also a few bugs and logic issues I flagged inline, take a look when you get a chance.

Can we grab some time to sync on this today? I want to walk through the comments together so we're on the same page about the direction, and then the fixes should be a lot more straightforward from there.

Requesting changes for now, but we'll sort it out soon :)

ColinToft:
@claude review

ludavidca (approved):
LGTM

ColinToft (approved):

PR #116, opened 2026-04-08
F4KRP-174 API endpoint to generate google maps link given a route_id

eddywang4340:
## JIRA Ticket
[Create a script + API endpoint inside of Python to generate a Google Maps link for drivers](https://f4kblueprint.atlassian.net/browse/F4KRP-174)

## Implementation Description

- Created a utils function that concatenates locations using their addresses (if available) and if not their coordinates
- Implemented a service that fetches the route and its locations and route_stops (from the route_id) and also fetches the warehouse info from system settings and then calls the utils function to create a google maps URL
- Added a new API route to get the google maps link from the route_service
- Added a demo script to demonstrate the result with the addition of a dummy route (will delete this script later)

## Steps to Test

1. Start your Docker containers
2. Run the demo script using this command: `docker-compose exec backend python scripts/demo_google_maps_link.py`

*

## Checklist


eddywang4340:
@claude review

eddywang4340:
Addressing Claude's first point about authentication for this google maps endpoint:

My initial thoughts are that no other routes in `route_routes.py` use authentication and the google-maps-link keeps that consistent. Claude does raise a good point though especially since this route exposes sensitive information like addresses, postal codes, geographical location, etc. @ludavidca do we have plans in the future to have authentication for these sensitive endpoints?

Either way I don't think this is in the scope of this ticket.

ludavidca:
Hey Eddy, Yes! @Jam-Cai is currently working on the authentication middleware ticket. Once that is finished, we will apply the authentication middleware to make sure this is only callable by drivers. Thanks for flagging this!

eddywang4340:
Addressing the rest of Claude's comments:
2. Implemented distinction between route not found and stops are not found
3. No tests will be added as part of this PR (this will be done in the future)
4. Implemented handling of max number of stops and char limit for URL
5. Demo script will be deleted before merging
6. Added newlines at the end of those files
7. Import ordering should be handled by lint
8. Added extra line between get_routes and new endpoint

ludavidca (approved):
Approved! Great Job as always, Eddy! Tested it locally and everything works swimmingly. 

Can you also update the seeding script so that there is a 10 location limit hardcoded (including warehouse as a stop)?


Small Food4thoughts:
1. To get around the 10 Locations Limit via Google Maps, one way we can skirt around this issue is by splitting it into 2 links and returning an array. If Emily or the drivers have an issue with the 10-location limit. This could be a solution.

PR #128, opened 2026-05-22
ci: always report required checks (fix stuck checks on path-filtered PRs)

ColinToft:
## Problem

`test` (Pytest CI) and `run-lint` (Lint codebase) are **required** status checks, but both were triggered by `paths:` filters (`backend/python/**`, and `frontend/**`+`backend/python/**` respectively). When a PR doesn't touch the filtered paths, the whole workflow is skipped, the required check never reports, and the PR is stuck on **"Expected — Waiting for status to be reported"** — unable to merge.

- A **frontend-only** PR (e.g. #127) skips `pytest.yml` → `test` hangs.
- A **CI/docs-only** PR (e.g. this one) skips `lint.yml` → `run-lint` hangs.

[Documented GitHub behavior](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks): a path-filtered *workflow* that's skipped reports **no** status (blocks a required check), whereas a *step/job* skipped via `if:` reports **success**. GitHub's guidance: don't use path filtering to skip a workflow that's required to merge.

## Fix

Drop the workflow-level `paths:` filters and gate the work at the **step** level instead (via `dorny/paths-filter`, which `lint.yml` already used):

- **`pytest.yml`** — run the `test` job on every PR; gate Python setup/install/pytest behind a `backend/python/**` filter.
- **`lint.yml`** — already gated every step behind the filter; just removed the workflow-level `paths:` so `run-lint` always runs and reports.

Result: both checks **always report a status**. On a no-op PR they skip the expensive steps and report success (no wasted backend test runs / installs); on a relevant PR they run exactly as before.

## Test plan

- [ ] This PR (CI-only) — both `test` and `run-lint` report success instead of hanging
- [ ] A backend PR still runs the full pytest suite + backend lint
- [ ] A frontend PR still runs frontend lint; `test` reports success without running backend tests

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,15 +4,9 @@ on:
   push:
     branches:
       - main
-    paths:
-      - "frontend/**"
-      - "backend/python/**"
   pull_request:
     branches:
       - main
-    paths:
-      - "frontend/**"
-      - "backend/python/**"
 
 jobs:
   run-lint:
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,13 +4,9 @@ on:
   push:
     branches:
       - main
-    paths:
-      - "backend/python/**"
   pull_request:
     branches:
       - main
-    paths:
-      - "backend/python/**"
 
 concurrency:
   group: pytest-${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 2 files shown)


ColinToft:
Folding these two CI commits into #127 instead — #127 is already a CI-hardening PR and was itself blocked by this path-filter bug, so carrying the fix in #127 lets it unblock itself (no merge-then-rebase dance).

PR #131, opened 2026-05-25
F4KRP-177: Implement Changes to Sweep Clustering Algorithm

Kerushani:
## JIRA Ticket
[F4KRP-177](https://f4kblueprint.atlassian.net/browse/F4KRP-177)

## Implementation Description
* This PR adds changes to the sweep clustering algorithm. Main changes are:
     * Changes how box counts are calculated: number of children / 2 + Round Up
     * Set a limit of 14 boxes max assigned to each driver
     * Determines far-away deliveries (using haversine to calculate its distance away from the warehouse, an estimated drive time, and route time budget) to give clusters with far-away deliveries fewer stops

* Locations are added to cluster by checking if they are first feasible (collected in `feasible_indices` and added via `_can_add_to_cluster` which checks if it violates box, stop and time constraint)
* Defines far away deliveries with `FAR_DISTANCE_KM_THRESHOLD` and `FAR_MAX_STOPS_PER_CLUSTER`.
* Far stop defined by two conditions (in `_is_far_location`):
     *  if address contains keywords outlined in `FAR_CITY_KEYWORDS`
     * OR if the location's distance is `>= FAR_DISTANCE_KM_THRESHOLD`
* Distance is calculated in `_haversine_km` using Haversine formula - determines the shortest great-circle distance between two points on a sphere using their latitudes and longitudes
     * Haversine formula is hav(φ) = sin^2(φ/2)
     * To avoid computing Haversine multiple times per location per cluster iteration, we cache the distance per location once at the start of the assignment group
          * `_LocationMetrics` holds `distance_km` (haversine),  `is_far`, and `stop_minutes`; the assignment loop only looks up `metrics_by_id[location.location_id]` — it does not recompute haversine per cluster try
     * There are comments in code to explain what's going on
* If a stop is far, max stops for that route is `min(max_stops, FAR_MAX_STOPS_PER_CLUSTER)`
* Routes with a far route are given a estimated time (`_far_route_minutes_cap`)
* `_estimated_minutes_for_stop` adds estimate minutes per service stop
* Sweep algo still uses squared differences in lat/lon not Haversine to order stops

(plots the warehouse, distance boundary, and far away locations on a graph)
Screenshot 2026-06-05 at 10 17 38 PM

(plot of clusters for +500 locations - it looks a bit crazy but the simple case of a small number of locations is fine)

Screenshot 2026-06-05 at 10 24 17 PM

The "X" within the boundary is from keyword cities
sweep_clustering_test

## Steps to Test

* Check if the test cases make sense
* Run `docker compose exec backend python scripts/sweep_algorithm_test.py` for sweep algo results
Screenshot 2026-06-07 at 9 24 49 PM
Screenshot 2026-06-07 at 9 25 58 PM
Screenshot 2026-06-07 at 9 26 22 PM

* Does the calculation for locations considered to be "far-away deliveries" (using location keyword, haversine calc, and route time estimation) make sense?
* There's `FAR_CITY_KEYWORDS` which hardcodes certain cities to be considered as "far-away", which is a quick initial check but there may be cases in which a street name for example, is in the `FAR_CITY_KEYWORDS` while its actually closeby. Is this okay / a case that probably won't come up too frequently?

## Checklist


backend/python/scripts/sweep_algorithm_test.py
```diff
@@ -2,109 +2,367 @@
...
 
-        if not cluster:
-            print("  (empty cluster)")
-            continue
+def _unique_hex_palette(count: int) -> list[str]:
+    """Return ``count`` visually distinct hex colours with no repeats."""
+    if count <= 0:
+        return []
 
+    golden_ratio = 0.618033988749895
```
ludavidca, line 66: Storing the golden ratio is crazy work lol!

ludavidca (approved):
LGTM

PR #139, opened 2026-05-30
Add Location Filters

aaronwxng:
Jira Ticket:
  - https://f4kblueprint.atlassian.net/browse/F4KRP-197?atlOrigin=eyJpIjoiM2VhMWEyZjE0YTBmNDQzN2FiY2RkODYxZjI5ODRhZTgiLCJwIjoiaiJ9 

Summary:
  - Add delivery type filters to `GET /locations` for School and Family locations
  - Add computed status filters to `GET /locations` for Active, Unscheduled, and Inactive
  - Add focused location route tests for delivery type and status filtering
  - Fix an existing syntax typo in `app/__init__.py` that prevented backend imports/tests from running


.postman/resources.yaml
```diff
@@ -0,0 +1,8 @@
+# Use this workspace to collaborate
```
ColinToft: We probably don't need to commit this, would you mind adding to the .gitignore:

```
.postman/
postman/globals/
```

backend/python/app/models/enum.py
```diff
@@ -36,6 +36,17 @@ class DeliveryTypeEnum(str, Enum):
     SUMMER = "Summer"
 
 
+class LocationDeliveryTypeEnum(str, Enum):
```
ColinToft: I think we can rename the values in the above enum (DeliveryTypeEnum) rather than creating a new enum for locations specifically. This way delivery types can be consistent everywhere in the platform

backend/python/app/services/implementations/location_service.py
```diff
@@ -92,16 +104,77 @@ async def get_location_by_id(
...
+            )
+            is_scheduled = scheduled_query.exists()
+
+            if status_filter:
+                status_conditions: list[Any] = []
+                state_column = cast("Any", Location.state)
+                is_active = state_column == LocationState.ACTIVE
+                is_archived = state_column == LocationState.ARCHIVED
+                if LocationStatusEnum.ACTIVE in status_filter:
+                    status_conditions.append(and_(is_active, is_scheduled))
```
ColinToft: OK so right now this filter is missing one combination of things: not is_active, but is scheduled. i.e. wasn't in the last import, but it is still scheduled in the future. I think we want to actually show this as "ACTIVE" still according to Isabelle. Also i realize this is pretty confusing to have a field called is_active that really means in_last_import, so as a side note I think we should probably rename this soon (in a separate ticket sometime).

Basically, I think this line can be status_conditions.append(is_scheduled) to just catch anything that is coming up in the future

backend/python/app/services/implementations/location_service.py
```diff
@@ -92,16 +104,77 @@ async def get_location_by_id(
...
         try:
             statement = (
                 select(Location)
-                .where(Location.state == LocationState.ACTIVE)
                 .options(selectinload(Location.location_group))  # type: ignore[arg-type]
                 .order_by(Location.created_at.desc())  # type: ignore[union-attr]
             )
+
+            if delivery_type:
+                school_name_column = cast("Any", Location.school_name)
```
ColinToft: I'm wondering if instead of the cast thing we could do stuff like :

```
col(Location.school_name).isnot(None)
col(Location.school_name) != ""
```

when we use the school name column?

The reason is that casting to "Any" basically is saying "just ignore the type of this". It does work, and gets around the `mypy` type checking errors. But if we do col() then it actually uses the correct type so that everything looks good to `mypy`, instead of having the types look a bit weird to `mypy` but relying on FastAPI's fancy stuff that makes the code work anyway.

backend/python/app/services/implementations/location_service.py
```diff
@@ -92,16 +104,77 @@ async def get_location_by_id(
...
+                is_archived = state_column == LocationState.ARCHIVED
+                if LocationStatusEnum.ACTIVE in status_filter:
+                    status_conditions.append(and_(is_active, is_scheduled))
+                if LocationStatusEnum.UNSCHEDULED in status_filter:
+                    status_conditions.append(and_(is_active, ~is_scheduled))
+                if LocationStatusEnum.INACTIVE in status_filter:
+                    status_conditions.append(and_(is_archived, ~is_scheduled))
+                if status_conditions:
+                    statement = statement.where(or_(*status_conditions))
+            else:
```
ColinToft: I think we can drop this else block. If we don't pass in any filters we don't need to filter on this.

(I think the previous code here did filter like this by default, but I think it's better, now that we have the filtering option available anyway, to just make it so that no filter = return everything)

ColinToft (changes requested):
Nice!! Thanks for implementing this. Just left a few comments with some changes to fix!

ColinToft (dismissed):

ColinToft (approved):
LGTM!

PR #143, opened 2026-06-01
Backend refactor: per-day routes + location lifecycle + snapshot tables

ColinToft:
## Summary

- **Locations** gain `in_roster` (stored "was in the latest spreadsheet upload" bit), replacing the old `state` ACTIVE/ARCHIVED enum. The UI-facing three-state `status` (Active/Unscheduled/Inactive) is **computed, not stored**: any present/future route → Active; otherwise in-roster → Unscheduled; otherwise → Inactive. Surfaced on `LocationRead` via a `computed_field` + a batched `has_future_route` lookup (no N+1).
  - `delivery_type` itself landed separately in #147 — this PR builds on it: `POST /locations/ingest` now takes `delivery_type` as a request-level parameter (one sheet = one type), replacing the hardcoded-Family TODO from #147.
- **Routes become per-day instances**: each `Route` belongs to exactly one `RouteGroup` via `Route.route_group_id`. The `RouteGroupMembership` M2M and the entire `DriverAssignment` table/service/router are dropped — driver assignment is now `Route.driver_id` (null = unassigned), plus `start_time` and `cloned_from_route_id` (lineage for the future Duplicate Route Group flow).
- **New snapshot tables** (`route_snapshots`, `route_stop_snapshots`) freeze delivery-relevant data (stop address/contact/phone/boxes/notes/coords + warehouse start) when a route's `drive_date` passes — extends the existing nightly driver-history cron, idempotently. Presence of a snapshot row == the route is frozen; no separate flag column. Until frozen, route stops read **live** from `Location`, so a family moving automatically updates upcoming routes while past routes stay historically accurate.
- **Driver suggestions** switch from "last driver on this route_id" to a location-familiarity heuristic (`get_suggested_drivers_for_route`): count of past frozen deliveries to the target route's locations, active drivers only.
- **Uniqueness guards on `route_stops`**: `UNIQUE(route_id, stop_number)` and `UNIQUE(route_id, location_id)` (a family can't be double-delivered within one route). `update_route` flushes stop deletions before re-inserting so reorders can't transiently violate them.
- `encoded_polyline`: `VARCHAR(10000)` → `TEXT` (the cap was a latent hard-failure on long routes). `routes.expires_at` dropped — `drive_date` carries the timeline.

**Migrations:** the main refactor migration is a **merge revision** joining main's two previously-unmerged heads (`4971451d79cd` user-invites, `e4f6a7b8c9d0` name/delivery_type) and inherits the `school_name` → `name` rename + `delivery_type` column from the latter. Existing routes/route_stops are deleted on migrate (re-seed required); locations are preserved with `state` → `in_roster` backfill. No prod data, so this is fine.

**Scope:** backend + the minimal frontend fallout (regenerated OpenAPI client; `LocationRead` → `LocationReadInput`/`LocationReadOutput` split from the computed field; Addresses tab Status column reads computed `status`). New UI features follow separately.

**Explicitly deferred** to follow-up PRs:
- Location similarity matcher / DB diff in `review_locations` (today's flow still has the client compute `net_new`/`stale`)
- "Changed/moved" ingest bucket
- `duplicate_route_group` endpoint + bulk-edit-forward over `cloned_from_route_id`
- Geographic / `location_group` cold-start fallback for the driver-suggestion heuristic

`cloned_from_route_id` is in place so when Duplicate ships, it's a pure service+endpoint add — no migration.

## Test plan

- [ ] Migration runs cleanly from a fresh DB: `alembic upgrade head` (single head after the merge revision)
- [ ] `python -m app.seed_database` succeeds and produces ~1100+ per-day route instances + snapshots for past routes
- [ ] `GET /locations` includes `in_roster`, `has_future_route`, and computed `status`; `?status=` and `?delivery_type=` filters work against the new semantics
- [ ] `GET /routes?unassigned_only=true` returns routes with `driver_id IS NULL`
- [ ] `GET /route-groups?include_routes=true` uses the direct `routes` relationship (no membership join)
- [ ] `GET /routes/{id}/google-maps-link` still works
- [ ] Mutating a location's address does not change a frozen past route's snapshot, but does propagate to upcoming routes (live FK)
- [ ] Inserting a duplicate stop_number or duplicate location on one route raises IntegrityError; PATCH reorder of stops still works (flush ordering)
- [ ] Run `ruff check`, `ruff format --check`, `mypy`, `pytest` — all green
- [ ] Verify nightly cron writes `RouteSnapshot` + `RouteStopSnapshot` rows for today's routes (run manually with a today override)


ColinToft:
@claude review

PR #144, opened 2026-06-02
Wilson/add aggregate response model to get /route groups

wilson-li3:
**## JIRA Ticket**
Ticket: https://f4kblueprint.atlassian.net/browse/F4KRP-196

**## Implementation Description**
****

- Updated GET /route-groups to return the aggregate fields needed
- Added num_locations, num_boxes, num_drivers_assigned, delivery_type, and status to RouteGroupRead with safe default values
- Updated the route group service to calculate these values with SQL subqueries
- Updated the router to include the calculated values in the API response

**## Steps to Test**
****
- Call GET /route-groups in Postman
-Verify route group includes num_locations, num_boxes, num_drivers_assigned, delivery_type and status
-I guess spot-check one route group manually to confirm the aggregate values look correct

**## What Should Reviewers Focus On?**
****
I guess just make sure none of the api request values look ridiculous, also maybe look at the sql query logic again


backend/python/app/services/implementations/route_group_service.py
```diff
@@ -75,9 +84,89 @@ async def get_route_groups(
...
+
+        now = datetime.now(self.timezone).replace(tzinfo=None)
+        today_start = now.replace(hour=0, minute=0, second=0, microsecond=0)
+        thirty_days_ago = now - timedelta(days=30)
+
+        status_expr = case(
+            (RouteGroup.drive_date > now, RouteStatusEnum.UPCOMING.value),
+            (RouteGroup.drive_date >= thirty_days_ago, RouteStatusEnum.COMPLETED.value),  # type: ignore[arg-type]
+            else_=RouteStatusEnum.ARCHIVED.value,
+        ).label("status")
```
ColinToft, line 164: I think we should use `today_start` here, not `now`. Reason being we want the platform treat anything today as upcoming, which basically means we mark routes as "completed" at the end of the day at midnight.

backend/python/app/services/implementations/route_group_service.py
```diff
@@ -22,6 +22,15 @@
...
 
+class RouteGroupWithStats(NamedTuple):
+    route_group: RouteGroup
+    num_locations: int
+    num_boxes: int
+    num_drivers_assigned: int
+    delivery_type: str | None
+    status: str
+
+
```
ColinToft: I'm a bit unsure about adding another type to the codebase, especially since I feel like this is duplicating a lot of the same stuff that `RouteGroupRead` has. (Also tbh I don't really know what a NamedTuple is although that isn't super relevant)

A couple other places in the code (see RouteService.get_routes, route_service.py:42-117) has the service build and return the Read model directly, by constructimg the read model in one call per row using the labeled Row (row.RouteGroup, row.num_locations). 

That way we can remove this NamedTuple and the data["..."] = row... loop in the router. The one confusing thing is the `include_routes` option, so I think we need a `routes` field added to RouteGroupRead as well so everything will fit.

backend/python/app/services/implementations/route_group_service.py
```diff
@@ -75,9 +84,89 @@ async def get_route_groups(
         route_status: list[RouteStatusEnum] | None = None,
         driver_assignment_status: list[DriverAssignmentStatusEnum] | None = None,
         include_routes: bool = False,
-    ) -> list[RouteGroup]:
-        """Get route groups with optional date filtering"""
-        statement = select(RouteGroup)
+    ) -> list[RouteGroupWithStats]:
```
ColinToft: Just adding to the comment from above, this is the function I think would be nice to have return `RouteGroupRead`

backend/python/app/routers/route_group_routes.py
```diff
@@ -58,10 +58,15 @@ async def get_route_groups(
             include_routes,
         )
         result = []
-        for route_group in route_groups:
+        for row in route_groups:
             data = RouteGroupRead.model_validate(
```
ColinToft: Also adding on to my other comment, I think if we have the service return `RouteGroupRead` we could remove some of this copying logic here

ColinToft (changes requested):
Good stuff!! Mostly just one change here (I think we can make the service return `RouteGroupRead` directly) and then I would also just make sure your PR passes linting , the README has more info on the formatting commands.

(I also added a test to check the edge case where the date is today and it should be considered upcoming, not complete. i think that is a one word fix in the code and then should be good to go)

ColinToft:
@claude review

backend/python/app/services/implementations/route_group_service.py
```diff
@@ -74,16 +79,99 @@ async def get_route_groups(
...
+            .label("num_boxes")
+        )
 
-        `delivery_type` filters on the per-Location delivery_type of a group's
-        stops; `driver_assignment_status` filters on whether a group's routes
-        have a driver_id set.
-        """
-        # Eager-load routes in the same query (one round-trip) rather than a
-        # per-group refresh below.
-        statement = select(RouteGroup).options(selectinload(RouteGroup.routes))  # type: ignore[arg-type]
```
ColinToft, line 86: Do you think it would be possible to bring this `selectinload` back? Or just do a `selectinload` or similar somewhere

backend/python/app/services/implementations/route_group_service.py
```diff
@@ -163,12 +248,47 @@ async def get_route_groups(
...
+        statement = statement.order_by(RouteGroup.drive_date)
 
         result = await session.execute(statement)
-        # routes are eager-loaded via selectinload on the statement, so they're
-        # safe to access (num_routes, route_group_routes.py) without lazy loads.
-        return list(result.scalars().all())
+        rows = result.all()
+
+        items: list[RouteGroupRead] = []
+        for row in rows:
```
ColinToft, line 257: Without the `selectinload`, in this loop each time we say `rg.routes` is actually going to be another query to the routes table.

Because we have a one-to-many relationship (one group has multiple routes) the best way to handle that in SQL is to have each row in the Route table store the ID of its group. 

So this for loop basically ends up making a bunch of queries, one for each row, something like

```
SELECT * FROM routes WHERE route_group_id = 'group-1';
SELECT * FROM routes WHERE route_group_id = 'group-2';
...
SELECT * FROM routes WHERE route_group_id = 'group-100';
```

whereas selectinload basically grabs all the routes at the same time as the route groups, which is better for when we know we need to return all the routes anyway. It basically does something like

`SELECT * FROM routes WHERE route_group_id IN ('group-1', 'group-2', ..., 'group-100');`

ColinToft (approved):
LGTM! One thing to fix and then good to merge

backend/python/app/services/implementations/route_group_service.py
```diff
@@ -74,16 +79,99 @@ async def get_route_groups(
...
+
+        now = datetime.now(self.timezone).replace(tzinfo=None)
+        today_start = now.replace(hour=0, minute=0, second=0, microsecond=0)
+        thirty_days_ago = now - timedelta(days=30)
+
+        status_expr = case(
+            (RouteGroup.drive_date >= today_start, RouteStatusEnum.UPCOMING.value),  # type: ignore[arg-type]
+            (RouteGroup.drive_date >= thirty_days_ago, RouteStatusEnum.COMPLETED.value),  # type: ignore[arg-type]
+            else_=RouteStatusEnum.ARCHIVED.value,
+        ).label("status")
```
ColinToft, line 164:
```suggestion
        status_expr = case(
            (RouteGroup.drive_date >= today_start, RouteStatusEnum.UPCOMING.value),  # type: ignore[arg-type]
            else_=RouteStatusEnum.COMPLETED.value,
        ).label("status")
```

ColinToft: We aren't gonna do the archiving for now in this platform I don't think, so we can remove the check here

backend/python/app/models/route_group.py
```diff
@@ -51,13 +51,28 @@ class RouteGroupCreate(RouteGroupBase):
...
 
     route_group_id: UUID
     created_at: datetime | None = None
     updated_at: datetime | None = None
     num_routes: int
+    num_locations: int = 0
+    num_boxes: int = 0
+    num_drivers_assigned: int = 0
+    delivery_type: str | None = None
+    status: str = "Archived"
```
ColinToft:
```suggestion
    delivery_type: DeliveryTypeEnum | None = None
    status: RouteStatusEnum
```

ColinToft: Let's use the enums here since we have them

backend/python/app/models/route_group.py
ColinToft, line 9:
```suggestion
from .enum import DeliveryTypeEnum, RouteStatusEnum

```

ColinToft (approved):
Couple more things but otherwise LGTM!

PR #145, opened 2026-06-02
Native frontend dev workflow for Windows; drop CHOKIDAR_USEPOLLING

ColinToft:
## Summary

- Document running the frontend natively on Windows as a workaround for the slow Docker bind-mount path
- Drop `CHOKIDAR_USEPOLLING=true` from `docker-compose.yml` — it was burning ~47% of a core continuously at idle for no benefit on Vite 7

## Why

Several of us have hit very slow frontend cold loads in Docker on Windows — minutes-long white screens that don't reproduce on macOS. Tried isolating it across:

| | cold `main.tsx` |
|---|---|
| Docker, HEAD, pnpm 9 (current) | ~80 s |
| Docker, HEAD, pnpm 11 | >60 s, wedged |
| Docker, commit from 5 weeks ago (`90c04d2`) | **179–184 s** |
| Docker, HEAD, with optimizeDeps + polling off | ~80 s first time, then warm-cached |
| **Native (no Docker), HEAD, pnpm 11** | **~9 s** |

A code-side regression doesn't fit — the 5-week-old commit is actually *slower* than HEAD. Pnpm version doesn't matter (tried 9 and 11). A Docker Desktop restart didn't help. The constant is the Windows bind-mount IO path, and the same code runs ~9× faster the moment you take Vite out of that path.

So the practical answer for now is: Windows devs run the frontend natively, keep the backend + db in Docker. Documenting that. Long-term fix (move repo into WSL2 filesystem) is mentioned but not blocking.

## Changes

- **`docker-compose.yml`**: remove `CHOKIDAR_USEPOLLING=true` from the frontend service. Measured idle CPU drop from ~47% → ~0.1%. Vite 7 doesn't need polling — chokidar's native fs events work fine.
- **`CONTRIBUTING.md`**: new section "Running the frontend on Windows" with the one-time setup, daily workflow, and a brief explanation.

## Notes for review

- @ludavidca — adding you because you ran into this earlier and tried `optimizeDeps.include` + lazy loading in #(your reverted PR). Worth knowing: removing polling alone didn't fix cold load in my testing (~80s either way), and `optimizeDeps.include` *with polling still on* actually wedged Vite (>240 s). So a future "fix it in Docker" attempt should drop polling first, then layer optimizeDeps. That said, this PR doesn't ship optimizeDeps because native dev is a much bigger win and avoids the trap.
- This change does not affect macOS / Linux dev workflows.

## Test plan

- [ ] On Windows: pull, `docker compose up backend db`, then in `frontend/` run `pnpm install && pnpm dev`. Confirm the page loads at http://localhost:3000 and API calls succeed against the backend container.
- [ ] On Windows: run `docker compose up` (full stack) and confirm the frontend container starts without `CHOKIDAR_USEPOLLING` and the page still renders (just slowly, as today).
- [ ] On macOS: `docker compose up` and confirm nothing regressed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,8 +106,7 @@ this is, but running locally is a good temporary fix.
 
 1. Install Node 24+ and pnpm 11+ on Windows (e.g. via [nvm-windows](https://github.com/coreybutler/nvm-windows) and `npm install -g pnpm`).
 2. From `frontend/`: `pnpm install`.
-
-No `.env` is required for native dev — the only frontend env var, `VITE_API_BASE_URL`, defaults to `http://localhost:8080` in code (`src/lib/axiosClient.ts`). Only create `frontend/.env` if you need to override that.
+3. From `frontend/`: `cp .env.example .env` (same `frontend/.env` Docker mounts via `env_file`). The default `VITE_API_BASE_URL=http://localhost:8080` points at the backend container's exposed port; override it if your backend runs elsewhere.
 
 ### Daily workflow
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,14 @@ docker-compose exec frontend pnpm lint:fix
 docker-compose exec frontend pnpm format
 ```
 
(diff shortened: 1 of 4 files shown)


ludavidca (approved):
LGTM

PR #147, opened 2026-06-06
Persist location delivery type

aaronwxng:
## Summary
- Rename location identity field from school_name to name and add stored delivery_type
- Update delivery type enum values to School / Family and filter route groups by Location.delivery_type
- Add Alembic migration, update seed/import/create paths, OpenAPI/generated types, and frontend filter labels
- Fix a pre-existing stray syntax token in backend app initialization that blocked imports/tests

## Tests
- docker compose exec -w /app backend python -m pytest
- docker compose run --rm frontend pnpm build


ColinToft:
@claude review

backend/python/app/models/location.py
```diff
@@ -24,8 +25,11 @@ class LocationBase(SQLModel):
     location_group_id: UUID = Field(
         foreign_key="location_groups.location_group_id", nullable=False
     )
-    school_name: str | None = None
+    name: str
     contact_name: str
+    delivery_type: DeliveryTypeEnum = Field(
+        default=DeliveryTypeEnum.FAMILY, sa_type=String
```
ColinToft: I think we can drop the default value here. All locations should have a delivery type and I think its good to be explicit about it (the default could maybe lead to weird silent failures)

ColinToft (approved):
LGTM! One small thing and then good to merge!

PR #155, opened 2026-06-11
Wilson/seed login usable firebase users

wilson-li3:
## JIRA Ticket

## Implementation Description

the seed script was making drivers with fake auth id values and the admin account depended on manually setting the admin auth env var. I updated the seed script so the seeded drivers and admins are real firebase users with testing credentials. running the seed is also idempodent and will not create duplicates everytime u run it

## Steps to Test

run docker-compose exec backend python -m app.seed_database from root, you should see it print out everything being created
run it again to ensure idempotent and see that it will not create duplicates
test it with postman using POST http://localhost:8080/auth/login and body: {"email": "driver001@f4k.dev", "password": "test123"}
it should show u the auth stuff
can also run same thing with admin body: {"email": "[admin1@f4k.dev](mailto:admin1@f4k.dev)", "password": "test123"}

*

## Checklist


backend/python/app/seed_database.py
```diff
@@ -605,12 +626,20 @@ def main() -> None:
...
+
+                ensure_firebase_user(
+                    uid=uid, email=email, password=SEED_PASSWORD, role="driver"
+                )
+
                 user = User(
-                    name=fake.name(),
-                    email=fake.email(),
-                    auth_id=f"seed_driver_{uuid.uuid4().hex[:8]}",
+                    name=f"Driver {n}",
```
ColinToft: Can we actually keep `fake.name()` for the name? I think it might be fun to have real names still instead of just "Driver 001"

backend/python/app/seed_database.py
```diff
@@ -857,26 +886,40 @@ def main() -> None:
...
+                admin_num = i + 1
+                uid = f"seed-admin-{admin_num}"
+                email = f"admin{admin_num}@f4k.dev"
+
+                ensure_firebase_user(
+                    uid=uid, email=email, password=SEED_PASSWORD, role="admin"
+                )
+
+                user = User(
+                    name=f"Admin {admin_num}",
```
ColinToft: Same thing here with the names

.env.example
```diff
@@ -7,8 +7,6 @@ DB_HOST=
 APP_ENV=
 APP_NAME=
 
-ADMIN_AUTH_ID=
```
ColinToft, line 10: Nice, good catch removing this!

ColinToft (approved):
LGTM!

PR #159, opened 2026-06-14
Add driver_id filter to GET /routes

ColinToft:
## What

Adds an optional `driver_id` query param to `GET /routes` so the driver homepage can fetch a single driver's routes in one request.

## Why

- The old `GET /driver-assignments/me` endpoint was removed by the refactor I did in #147 (so there was no way to get a driver's routes) and this replaces it.

## Changes

- `driver_id` threads `route_routes.get_routes` → `route_service.get_routes`, adding `WHERE Route.driver_id == driver_id`.
- Combining `driver_id` with `unassigned_only` returns **400** — a route can't be both unassigned and assigned to a specific driver (fail fast rather than silently returning nothing).
- Tests: filter returns only that driver's routes (excludes other drivers + unassigned), empty page for a driver with none, and the 400 guard.
- Regenerated `frontend/openapi.json` + TS client for the new param.

### DIFF
--- a/backend/python/app/dependencies/auth.py
+++ b/backend/python/app/dependencies/auth.py
@@ -228,26 +228,36 @@ async def resolve_route_list_driver_filter(
     access_token: str = Depends(get_access_token),
     session: AsyncSession = Depends(get_session),
 ) -> UUID | None:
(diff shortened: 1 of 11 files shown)


ColinToft:
@claude review

PR #160, opened 2026-06-15
schema-changes

aaronwxng:
This updates the way we store people and location contact info so the app can use first and last names, show a full name when needed, and support a second phone number for locations. I also updated the seed data, import flow, and tests so everything still works together.\n\nI checked it by running the backend test suite and the frontend build in Docker.


.postman/resources.yaml
```diff
@@ -0,0 +1,8 @@
+# Use this workspace to collaborate
+workspace:
+  id: 0e5da946-3ec2-40bc-8da1-73672c2d7dae
+
+localResources:
+  docs:
+    - ../frontend/README.md
+services/jobs/README.md
```
ColinToft: Would you mind removing the postman stuff? And maybe look into why these are getting added, they are in the .gitignore so they shouldn't be in the repo unless something is like force-adding them?

ColinToft:
For the new nullable fields: `partner_driver_name` (and `phone_secondary` on Location) are the first genuinely-nullable columns in these models, but the update paths use `if field is not None: setattr(...)`. That means once one is set, there’s no way to clear it back to null.


Claude's suggestion is to use `model_dump(exclude_unset=True)` and iterating through the set fields, so omitted keys are skipped but an explicit `null` clears the value.

```
update_data = driver_data.model_dump(exclude_unset=True)
for field, value in update_data.items():
    setattr(driver, field, value)
```

Because of this we'll also have to adjust the manual `old_*` rollback block in `update_driver_by_id`.

ColinToft (approved):
LGTM!!

PR #142, opened 2026-05-31
Scheduled and Event-Driven Email-Sending

ChloeW125:
## JIRA Ticket
[Backend Jobs Component for Email Flow](https://f4kblueprint.atlassian.net/jira/software/projects/F4KRP/boards/1?jql=assignee%20%3D%20712020%3Ad9f87391-0501-4d3f-bfa6-8ea683925cb5&selectedIssue=F4KRP-201)

## Implementation Description

* Implemented the backend email dispatch pipeline to support scheduled and event-driven email sending.
* Updated frontend email templates to accomodate for Jinja2-based email variable handling (not using regex to avoid code injection security concerns)

## Steps to Test

1. Hit the test endpoint with any event-driven email of your choosing and see if an email gets sent/can be received (drivers/test-event-email, will need to also go into code to modify what info you want the email fields to be populated with in the sent email)
2. Schedule the route reminder email (scheduler-based email) and see if it gets sent at the right time

Note: Image assets will not render in email; this is because right now we don't have anywhere to host them; this shouldn't be a problem once we have that set up.

## What Should Reviewers Focus On?

* Edge cases!
* Slip-ups in formatting or styling that should be addressed in the actual sent email (note: please ignore discrepancies in formatting relating to the images or font, I will open another PR to fix those)

## Checklist
- [x] PR name and commits are descriptive, imperative, and atomic (trivial commits squashed)
- [x] I have requested a review from Claude, understood its suggestions, and implemented fixes where needed (or documented disagreements)
- [x] I have requested a review from the PL and relevant devs with background on this PR

PR #161, opened 2026-06-16
Settings backend (F4KRP-206): global settings, per-notification email reminders, PATCH endpoint

aaronwxng:
This updates the settings backend so admins can edit the new global values from ticket 206. It adds the missing write endpoint, stores the new settings fields, updates the reminder email job to use the configured schedule, and refreshes the API client.

I also updated the seed data and added coverage for the new settings path.

Validation:
- 205 backend tests passed in Docker
- frontend build passed in Docker
- OpenAPI client regenerated


ColinToft:
@claude review

.postman/resources.yaml
```diff
@@ -0,0 +1,8 @@
+# Use this workspace to collaborate
+workspace:
+  id: 0e5da946-3ec2-40bc-8da1-73672c2d7dae
+
+localResources:
+  docs:
+    - ../frontend/README.md
+services/jobs/README.md
```
ColinToft: Would you mind removing the postman files from this (and future PRs)?

backend/python/app/services/jobs/email_reminder_jobs.py
```diff
@@ -0,0 +1,122 @@
+"""Email reminder scheduled jobs"""
+
```
ColinToft: OK this is a bit unfortunate but I guess there was some bad timing in when the email logic was being worked on. Chloe recently merged #142 which moved the email logic into email_jobs.py (which is why it looks like this PR is adding this `email_reminder_jobs.py` file fresh). Do you think you'd be able to make the email changes on `email_jobs.py` instead and then delete this file

ColinToft (dismissed):
Some unfortunate timing with another recent change, so I was wondering if you could redo the changes to the email service on the new email_jobs.py file? Thanks so much!

ColinToft (approved):
LGTM!

PR #162, opened 2026-06-16
route list counts

aaronwxng:
Adds per-route stop and box totals to GET /routes so the route list can show summary counts without loading every stop.

Validated with backend route/auth integration tests, Ruff, and OpenAPI sync.

### DIFF
--- a/backend/python/app/services/implementations/route_service.py
+++ b/backend/python/app/services/implementations/route_service.py
@@ -3,7 +3,7 @@ from datetime import datetime, timezone
 from uuid import UUID
 
 from fastapi import HTTPException
-from sqlalchemy import func
+from sqlalchemy import Integer, case, cast, func
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlmodel import col, select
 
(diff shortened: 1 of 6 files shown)


ColinToft:
@claude review

backend/python/app/services/implementations/route_service.py
```diff
@@ -57,10 +57,29 @@ async def get_routes(
         to routes assigned to that specific driver (powers the driver homepage
         feed). The date range filters on the route's RouteGroup.drive_date.
         """
-        statement = select(Route, RouteGroup.drive_date).join(
+        route_totals = (
+            select(
+                col(RouteStop.route_id).label("route_id"),
+                func.count(col(RouteStop.route_stop_id)).label("num_stops"),
+                func.coalesce(func.sum(col(Location.num_boxes)), 0).label("box_total"),
```
ColinToft: OK a couple quick things here:
1. For routes that are already completed (day < today) we store a snapshotted version. This is so we can edit locations and move them around without messing up completed routes. So can we change this query so that for routes in the past can we read the value from RouteStopSnapshot instead?
2. Sorry this was not clear from the ticket (my bad) but `num_boxes` should be computed, as `ceil(num_children / 2)`. I thought this column was already gone, I'm making a quick change to remove this rn. If #161 happens to merge first you can use the `num_children_per_box` setting instead of hardcoding 2, but either is fine!

ColinToft (approved):
2 things which would be nice to fix, then LGTM!

PR #163, opened 2026-06-17
Driver monthly kms and deliveries

ChloeW125:
## JIRA Ticket
[Ticket Name](https://f4kblueprint.atlassian.net/jira/software/projects/F4KRP/boards/1?jql=assignee%20%3D%20712020%3Ad9f87391-0501-4d3f-bfa6-8ea683925cb5&selectedIssue=F4KRP-146)

## Implementation Description

3 API Endpoints:
1. Total deliveries between a specified start and end date (dates defaulted to being treated in EST)
2. Monthly driver ranking by kms driven during that month
3. Monthly total number of kms driven and deliveries made

Notes:
* Assume all dates are treated in EST, unless specified otherwise
* Slotted endpoints under "reports" because this data is all aggregate data (i.e. this data is aggregated from all drivers' data), which differs from the driver/route/location, etc.-specific data extracted via other endpoints

## Steps to Test

* Make sure hitting the endpoints works! (Don't return a 403 or 500, etc.)
* Double-check output amounts

* Make sure all fields needed are included (and that all fields that need to be marked required are marked accordingly)
* Make sure response body matches expectations
* Verify it works!

## Checklist


ColinToft:
@claude review

backend/python/app/routers/report_routes.py
```diff
@@ -0,0 +1,104 @@
...
+    return dt.astimezone(tz)
+
+
+@router.get("/deliveries/count")
+async def get_total_deliveries_between(
+    start: datetime = Query(..., description="Start datetime (assumed EST if no tz)"),
+    end: datetime = Query(..., description="End datetime (assumed EST if no tz)"),
+    session: AsyncSession = Depends(get_session),
+    _auth: bool = Depends(require_admin),
+) -> dict[str, int]:
```
ColinToft: I do kinda like reviewer claude's suggestion of adding types to these responses (will also be nice for the frontend to have typed things)

backend/python/app/services/implementations/driver_report_service.py
```diff
@@ -0,0 +1,120 @@
...
+        self.logger = logger
+        self.timezone = ZoneInfo(settings.scheduler_timezone)
+
+    async def get_monthly_km_ranking(
+        self, session: AsyncSession, year: int, month: int
+    ) -> list[dict]:
+        """Return per-driver km for given year/month ordered desc by km."""
+        try:
+            # Build explicit ON clauses but cast their type for mypy
+            on_driver = cast("Any", Driver.driver_id == DriverHistory.driver_id)
```
ColinToft: I kind of don't like casting to Any type because it loses type information. You can do `col(<some column name>)` which helps get the actual type correct. Like I think if you do `col(Driver.driver_id) == col(DriverHistory.driver_id)` you wouldn't need to cast, and you could probably put this chunk back inside the original query also if you wanted

ColinToft (approved):
Couple small changes, but otherwise LGTM and feel free to merge after!

PR #166, opened 2026-06-19
Derive box counts from num_children; drop stored num_boxes

ColinToft:
Stacked on #161 (base = `f4krp-206-settings-backend`), since this depends on the `children_per_box` setting that #161 adds. Retarget to `main` once #161 merges.

## What & why
Box counts were a stored `Location.num_boxes` field, which let the stored value drift from the real driver (`num_children`). This removes the stored field and derives boxes everywhere as `ceil(num_children / children_per_box)` via a single helper (`app/utilities/boxes.compute_boxes`), using the configurable `children_per_box` setting instead of the old hardcoded `/2`.

## Changes
- **`Location`**: `num_children` now required (`ge=0`); `num_boxes` removed.
- **`RouteStopSnapshot`**: stores `num_children` instead of `num_boxes`.
- **`children_per_box`**: tightened to `ge>=1` (it's a divisor).
- **Clustering stack** (sweep / k-means / mock): `children_per_box` threaded through so the divisor comes from settings, not a constant. The hardcoded `/2` in `effective_boxes` is gone.
- **Live consumers**: snapshot job, location import, and seed updated to children.
- **Import**: now expects a **"Number of Children"** column, and the frontend ImportStep field key is `num_children` (was `num_boxes` — would have left every imported location with 0 children).
- **Migration**: backfills null `num_children`→0 before `NOT NULL`, drops `num_boxes` on both tables.
- **OpenAPI**: regenerated `openapi.json` + TS client.

## Backfill note
Existing rows with no children data backfill to `0` (the old box count can't be inverted to a children count).

## Validation
- Backend: `pytest` **283 passed**; `alembic upgrade head` + `alembic check` clean (no drift); mypy + ruff clean.
- Frontend: `tsc -b`, `eslint`, `prettier --check` all clean.

## Follow-up (separate PR)
#162 ("route list counts") sums `Location.num_boxes` in SQL — whichever of these merges second must switch its `box_total` aggregate to per-location `ceil(num_children / children_per_box)`. Not changed here to avoid touching that PR.


ColinToft:
@claude review

PR #171, opened 2026-06-22
Wilson/addresses table backend aggregates

wilson-li3:
## JIRA Ticket

## Implementation Description

*
i added three read only fields to LocationRead: assigned_route, last_delivery_date, and total_deliveries. These are derived from route data at query time

## Steps to Test

use postman to hit http://localhost:8080/locations to see the new endpoint
you will need an auth bearer token

*
should double check the two new SQL query methods in location_service.py which are load_assigned_routes and load_delivery_aggregates and make sure the join logic and filtering is right

## Checklist


ColinToft (approved):
LGTM! It would be nice if you could add a test for this new functionality but otherwise good to merge

PR #172, opened 2026-06-23
schema-changes

aaronwxng:
This updates announcement edit/delete so only the original author or an admin can change them. It keeps the existing driver self-edit behaviour in place and updates the auth tests to cover the ownership rules.\n\nValidated with targeted backend tests in Docker: PYTHONPATH=/app pytest tests/test_auth_integration.py tests/test_routes.py -q


ColinToft (approved):
LGTM!!

PR #174, opened 2026-06-23
Add dev-only design-overlay comparison tool

ColinToft:
## Summary

Adds a dev-only **design-overlay** tool — a lightweight, in-app "PerfectPixel" — for checking the running app against Figma exports. It renders a chosen image as a fixed, opacity-adjustable layer over the app so you can eyeball pixel alignment against the design.

Motivation: comparing implemented screens (e.g. the announcement board) against Figma frames currently means eyeballing side-by-side screenshots. This lets you lay the design directly over the live app.

## What it does
- **Load an image at runtime** — file picker, drag-drop onto the panel, or paste (⌘V). No design PNGs are committed to the repo.
- **Controls** — opacity slider, scale, X/Y offset (arrow-key nudge while the panel is focused; Shift = 10px), fit-width, reset, and a toggle hotkey (**Cmd/Ctrl+Shift+O**).
- **Non-intrusive** — the image layer is `pointer-events: none`, so the app stays fully usable underneath. Settings persist in `localStorage`.
- **Dev-only** — mounted behind `import.meta.env.DEV` in `App.tsx`; the constant folds to `false` in prod so it's tree-shaken out.

## Implementation notes
- Single self-contained component (`frontend/src/dev/DesignOverlay.tsx`) using **inline styles**, so it can't touch the app's design-system tokens or trip the Tailwind/breakpoint lint rules.
- Clipboard-paste and drag-drop are scoped so they don't hijack the app's own inputs/DnD (paste ignored while typing in a field; drop only on the panel).

## Testing
- `eslint src` (full) — clean; `prettier --check` — clean; `tsc -b` — clean; `npm run build` — succeeds.
- Verified live against the running stack: control panel renders, image overlays at the set opacity/offset, `pointer-events: none` confirmed, app interactive underneath.

## Screenshot
Overlay (pink "DESIGN FRAME" stand-in) layered over the admin home at 55% opacity, with the control panel:

_(see PR thread)_

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -6,6 +6,7 @@ import {
   NotFoundPage,
   ServiceUnavailablePage,
 } from './common/components';
+import { DesignOverlay } from './dev/DesignOverlay';
 import { AdminLayout, DriverLayout } from './layouts';
 import {
   AdminDriversPage,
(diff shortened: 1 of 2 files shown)


ColinToft:
Keeping this local for now — not merging.

PR #175, opened 2026-06-26
Add location group filter to locations

aaronwxng:
## What changed
This adds a new `location_group_id` filter to `GET /locations/`, so callers can fetch locations for one or more location groups from the normal paginated locations endpoint.

I also added tests for filtering by one group, multiple groups, using it with the existing delivery type and status filters, and making sure pagination still works.

## Tested
- `docker compose exec backend ruff check app tests`
- `docker compose exec backend mypy . --config-file mypy.ini`
- `docker compose exec backend python -m pytest tests/test_routes.py`
- `docker compose exec backend python -m pytest`


ColinToft (approved):
LGTM!!

PR #176, opened 2026-06-27
F4KRP-233 extended get routes with route_id to include ordered stops in response

eddywang4340:
## JIRA Ticket
[Extend GET /routes/{id} to include the route's ordered stops](https://f4kblueprint.atlassian.net/browse/F4KRP-233)

## Implementation Description

- Added RouteDetailRead response to route.py (so that response includes details about the route_stops)
- Added RouteStopDetailRead response to route_stop (so that stops include all details that were missing from RouteStopRead, which only included the stop id)
- Modifed route_routes so that the GET routes endpoint returns the new RouteDetailRead instead of old Route
- Modified route_service: created helper called fetch_ordered_stops (does exactly as the name suggests) and used said helper in get_routes to get the ordered stops so that we can return it in the response to the GET endpoint
- Modified and added tests to test_routes (added tests for testing that if a snapshot exists we use the snapshot values in the GET endpoint)
- Frontend changes are purely generated after switching to new type for the GET endpoint

## Steps to Test

1. You can test this using primarily endpoints from /docs
2. First, make sure you auth by hitting the /auth/login endpoint (using an admin email and password) to obtain an access token
3. Scroll to the top of the docs and paste in your access token after clicking Authorize
4. Find a route_id that you would like to GET
5. Use that route_id to obtain a route and make sure you get details about the route stops

*

## Checklist


eddywang4340:
Addressing Claude's comments:
1. Addressed this by removing the call to get_routes and instead just checking if that route exists so we can determine which error to propogate
2. This is an edge case that theoretically should never happen so this comment was ignored

ColinToft (approved):
LGTM!!

PR #181, opened 2026-06-30
Make delivery types configurable

aaronwxng:
## Summary
- Moves delivery types into system settings, seeded with the current School and Family values.
- Validates location create, update, and ingest requests against the configured delivery types.
- Updates route/address filters to read delivery types from settings instead of using hardcoded values.

## Testing
- docker compose exec backend ruff format --check app tests scripts
- docker compose exec backend python -m pytest
- corepack pnpm@9.15.9 lint
- corepack pnpm@9.15.9 build
- docker compose exec backend alembic heads
- git diff --check


ColinToft:
@claude review

backend/python/app/routers/location_routes.py
```diff
@@ -101,6 +111,11 @@ async def create_location(
     try:
         created_location = await location_service.create_location(session, location)
         return LocationRead.model_validate(created_location)
+    except InvalidDeliveryTypeError as ve:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
ColinToft, line 116: I see here you are using 400 error code if the delivery type was invalid, but above uses a 422.  The specific error code isn't super important just wondering what the reasoning was to make them different?

backend/python/app/services/implementations/route_group_service.py
```diff
@@ -127,30 +126,27 @@ async def get_route_groups(
...
             .correlate(RouteGroup)
+            .scalar_subquery()
         )
 
         delivery_type_expr = case(
-            (has_school_subq.exists(), DeliveryTypeEnum.SCHOOL.value),
-            (has_locations_subq.exists(), DeliveryTypeEnum.FAMILY.value),
-            else_=None,
+            (school_delivery_exists, "School"),
+            else_=delivery_type_fallback,
```
ColinToft: Can we rewrite this part without hardcoding "School" as a type here, and probably without the `has_school_subq` as well? Ideally we don't want to have behaviour depend on any specific type, so the client is free to name them whatever. 

I think what was happening here was that the previous query was trying to handle route groups that were a mix of family and school, and defaulting them to school if any location in there was a school. But that was not actually necessary because everything in a route group will all have the same delivery type (this is an **invariant** meaning we are enforcing that rule throughout the program). So you could just look at any location in the route group to find the delivery type of that whole route group.

backend/python/app/services/implementations/location_service.py
```diff
@@ -236,6 +240,39 @@ async def get_locations(
             self.logger.error(f"Failed to get locations: {e!s}")
             raise e
 
+    async def get_delivery_types(self, session: AsyncSession) -> list[str]:
+        """Return configured delivery types, falling back to current defaults."""
+        settings = await self.system_settings_service.get_settings(session)
+        if settings is None:
```
ColinToft: I think we can remove this check for None/fallback to default. Ideally, I'd want settings to never be None. (I just added in a quick PR #186 that makes sure settings always exists on startup so now we are definitely good to remove this check).

frontend/src/api/system-settings.ts
```diff
@@ -1,6 +1,17 @@
 import { useQuery } from '@tanstack/react-query';
 
 import { getSystemSettingsOptions } from './generated/@tanstack/react-query.gen';
+import type { SystemSettingsRead } from './generated/types.gen';
+
+const DEFAULT_DELIVERY_TYPES = ['School', 'Family'];
```
ColinToft: Is it possible to remove this default value here? similar to before since the goal is to make delivery types customizable I want to try and avoid hardcoding the old "school" and "family" options. now that the settings row should always exist we shouldn't need the default.

ColinToft (changes requested):
Nice work! Just a couple comments and one clarification question.

ColinToft (approved):
LGTM, thanks for fixing those comments!

PR #182, opened 2026-07-03
Implement import merging logic

aaronwxng:
## Summary
- Adds import review classification for net-new, stale, and changed locations using the 2-of-3 name/address/phone matching rule.
- Splits import validation into specific blocking alert codes and returns duplicate groups with every duplicate row included.
- Updates ingest to apply stale rows, non-address changes, and address-change replacements while preserving notes/history.
- Regenerates the OpenAPI client and updates the validation screen for the new alert codes.

## Testing
- docker compose exec backend python -m pytest tests/test_routes.py::TestLocationImportRoutes -q
- docker compose exec backend python -m pytest tests/test_routes.py
- docker compose exec backend python -m pytest
- docker compose exec backend ruff format --check app tests scripts
- corepack pnpm@9.15.9 lint
- corepack pnpm@9.15.9 build
- git diff --check


ColinToft:
@claude review

ColinToft:
@claude review

backend/python/app/models/location.py
```diff
@@ -79,16 +79,21 @@ def location_group_name(self) -> str:
...
     MISSING_PHONE_NUMBER = "MISSING_PHONE_NUMBER"
     INVALID_PHONE_NUMBER = "INVALID_PHONE_NUMBER"
-    MISSING_SCHOOL_OR_LAST_NAME = "MISSING_SCHOOL_OR_LAST_NAME"
-    INVALID_SCHOOL_OR_LAST_NAME = "INVALID_SCHOOL_OR_LAST_NAME"
+    MISSING_NAME = "MISSING_NAME"
+    INVALID_NAME = "INVALID_NAME"
     MISSING_DELIVERY_GROUP = "MISSING_DELIVERY_GROUP"
-    DUPLICATE_ENTRY = "DUPLICATE_ENTRY"
+    LOCAL_DUPLICATE = "LOCAL_DUPLICATE"
+    # Deprecated import alert codes kept for old clients that may still decode
```
ColinToft: We don't need to deprecate these alert codes, I think we can just remove them (and all references to them elsewhere in the code).

This is something to watch out for when using AI: it tends to think a bit too hard about backwards-compatibility, keeping support for old code, and it won't just fully change things over. This can be sometimes helpful but in this case it's not needed since we don't actually have anyone using the platform yet (so no "old clients" to think about)

frontend/src/api/generated/types.gen.ts
```diff
@@ -7,17 +7,20 @@ export type ClientOptions = {
...
   | 'MISSING_PHONE_NUMBER'
   | 'INVALID_PHONE_NUMBER'
-  | 'MISSING_SCHOOL_OR_LAST_NAME'
-  | 'INVALID_SCHOOL_OR_LAST_NAME'
+  | 'MISSING_NAME'
+  | 'INVALID_NAME'
   | 'MISSING_DELIVERY_GROUP'
-  | 'DUPLICATE_ENTRY';
+  | 'LOCAL_DUPLICATE'
+  | 'MISSING_FIELDS'
```
ColinToft: Can also remove these here (as opposed to deprecating, see other comment)

frontend/src/pages/admin/routes/generation/ValidateStep.tsx
```diff
@@ -43,8 +43,14 @@ function getAlertDisplay(code: AlertCode): {
       return { type: 'error', label: 'Invalid Phone Number' };
     case 'MISSING_DELIVERY_GROUP':
       return { type: 'error', label: 'Missing Delivery Group' };
-    case 'DUPLICATE_ENTRY':
+    case 'LOCAL_DUPLICATE':
       return { type: 'error', label: 'Duplicate Entry' };
+    case 'MISSING_FIELDS':
```
ColinToft: Think we can remove these as well

backend/python/app/models/location.py
```diff
@@ -125,6 +130,12 @@ class LocationImportRow(SQLModel):
     alerts: list[AlertCode]
 
 
+class DuplicateGroup(SQLModel):
+    """Rows that refer to the same imported location under the 2-of-3 rule."""
+
+    rows: list[int]
```
ColinToft, line 138: i'm wondering if we could actually have a way to flag which fields are matching so that we can display that on the frontend (designs have the specific fields highlighted in red)

ColinToft (approved):
LGTM! couple comments and then good to merge!

PR #184, opened 2026-07-06
F4KRP 208: Align location import validation with Current Validate Screen Requirements

Kerushani:
## JIRA Ticket
[F4KRP 208](https://f4kblueprint.atlassian.net/browse/F4KRP-208)

## Implementation Description

- Updates POST /locations/review to align with the Figma Validate screen and import PRD. Validation logic lives in `location_import_validation.py`; `LocationService.review_locations()` orchestrates parsing, geocoding, and response assembly.

- Per-field blocking errors -> Replaced lumped `MISSING_FIELDS` / `INVALID_FORMAT` with specific codes. `collect_field_alerts()` evaluates each field independently, so one row can report multiple errors (e.g. `INVALID_SCHOOL_OR_LAST_NAME` + `MISSING_DELIVERY_GROUP`).

- Small update to frontend to use new AlertCodes

New validations:
- Invalid address -> `review_locations()` geocodes via `_geocode_import_address()`; `collect_field_alerts()` emits `INVALID_ADDRESS` when `geocode_ok` is False.
     - Invalid name -> `is_invalid_school_or_last_name()` flags all-digit names (e.g. 33333333).
     - Invalid phone -> `try_normalize_phone()` normalizes valid numbers to E.164; invalid non-empty values get `INVALID_PHONE_NUMBER`.
- Duplicate detection (2-of-3 rule) -> Two rows are linked if `rows_are_duplicates()` returns true, which delegates to `count_duplicate_field_matches()` and requires ≥ 2 matches across {name, address, phone}. Matching uses `_name_match_key()` (case-insensitive), `_address_match_key() `(exact stripped string; 123 Main St ≠ 123 Main Street), and `_phone_match_key()`. Removed the old address-or-phone-only check that over-flagged shared buildings.

- **Transitive grouping (using union-find to group duplicates)**

Pairs alone aren't enough; duplicates can chain transitively:

| Row | Name | Address | Phone | Linked to |
|---|---|---|---|---|
| A | Alpha | 1 A St | 519-111-1111 | B (name + phone) |
| B | Alpha | 2 B St | 519-111-1111 | A (name + phone), C (address + phone) |
| C | Gamma | 2 B St | 519-111-1111 | B (address + phone) |

A and C share only phone (1-of-3) -> `rows_are_duplicates(A, C)` is `false`, but all three belong in one group because B connects them.

**How `find_duplicate_index_groups()` handles this**

1. Compare every row pair; if `rows_are_duplicates()` → `_UnionFind.union(i, j)`
2. Collect connected components (size ≥ 2) as 0-based index groups
3. `review_locations()` maps indices → 1-based row numbers in `duplicate_groups`
4. Every row in a cluster gets `DUPLICATE_ENTRY` appended to its alerts

## Steps to Test

1. Unit tests added for location validation logic

* Invalid name ->` is_invalid_school_or_last_name()` flags all-digit names (e.g. 33333333).
    * Is there any other criteria for an invalid school or last name?
* Does the grouping make sense? Should `A` and `C` be grouped together on the UI even though they are not duplicates?

## Checklist

**Frontend only — delete if not applicable:** -> frontend changes are small
- [x] Follows Tailwind/shadcn best practices per README; no hardcoded color or spacing values
- [x] Implementation matches Zeplin designs and screenshots are attached above
- [x] Screens render correctly on mobile and desktop; no console warnings or errors

### DIFF
--- a/backend/python/app/services/implementations/location_import_validation.py
+++ b/backend/python/app/services/implementations/location_import_validation.py
@@ -44,13 +44,14 @@ def collect_field_alerts(
     *,
     geocode_ok: bool | None,
     phone_invalid: bool,
+    phone_secondary_invalid: bool = False,
 ) -> list[AlertCode]:
     """Collect per-field blocking alerts for one parsed import row."""
     alerts: list[AlertCode] = []
 
     if is_blank(entry.contact_name):
         alerts.append(AlertCode.MISSING_SCHOOL_OR_LAST_NAME)
-    elif present_str(entry.contact_name) and is_invalid_school_or_last_name(
+    elif entry.contact_name is not None and is_invalid_school_or_last_name(
         entry.contact_name
     ):
         alerts.append(AlertCode.INVALID_SCHOOL_OR_LAST_NAME)
@@ -63,6 +64,9 @@ def collect_field_alerts(
     if is_blank(entry.phone_primary):
         alerts.append(AlertCode.MISSING_PHONE_NUMBER)
     elif phone_invalid:
+        alerts.append(AlertCode.INVALID_PHONE_NUMBER)
+
+    if phone_secondary_invalid:
         alerts.append(AlertCode.INVALID_PHONE_NUMBER)
 
     if is_blank(entry.delivery_group):
@@ -80,7 +84,7 @@ def _name_match_key(name: str | None) -> str | None:
 def _address_match_key(address: str | None) -> str | None:
     if not present_str(address):
         return None
-    return address.strip()
+    return address.strip().casefold()
 
 
 def _phone_match_key(phone: str | None, normalized_phone: str | None) -> str | None:
--- a/backend/python/app/services/implementations/location_import_validation.py
+++ b/backend/python/app/services/implementations/location_import_validation.py
@@ -2,10 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, TypeGuard
-
-if TYPE_CHECKING:
-    from collections.abc import Sequence
+from typing import TypeGuard
 
 from app.models.location import AlertCode, LocationImportEntry
 from app.utilities.utils import validate_phone
@@ -87,9 +84,12 @@ def _address_match_key(address: str | None) -> str | None:
     return address.strip().casefold()
 
 
-def _phone_match_key(phone: str | None, normalized_phone: str | None) -> str | None:
-    if normalized_phone:
-        return normalized_phone
+def _phone_match_key(phone: str | None) -> str | None:
+    """Match key for phone_primary.
+
+    Callers should normalize phones onto the entry before duplicate detection
+    (see review_locations) so valid numbers compare as E.164.
+    """
     if not present_str(phone):
         return None
     return phone.strip()
@@ -98,11 +98,8 @@ def _phone_match_key(phone: str | None, normalized_phone: str | None) -> str | N
 def count_duplicate_field_matches(
     left: LocationImportEntry,
     right: LocationImportEntry,
-    *,
-    left_phone: str | None,
-    right_phone: str | None,
 ) -> int:
-    """Count how many of {name, address, phone} match between two rows, which is the 2-of-3 rule"""
+    """Count how many of {name, address, phone} match between two rows (2-of-3 rule)."""
     matches = 0
 
     left_name = _name_match_key(left.contact_name)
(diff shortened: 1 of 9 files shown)


ColinToft:
@claude review

frontend/src/pages/admin/routes/generation/ValidateStep.tsx
```diff
@@ -95,7 +109,7 @@ export function ValidateStep() {
       render: (row) => {
         const first = row.alerts[0];
```
ColinToft: I guess this was already in the existing code, but is it possible to show multiple alerts if they exist, like in the Figma:

Image

frontend/src/pages/admin/routes/generation/ValidateStep.tsx
ColinToft: I think we decided everything is just gonna be red and needs to be fixed (no separate errors or warnings)

backend/python/app/services/implementations/location_import_validation.py
```diff
@@ -0,0 +1,192 @@
...
+            left,
+            right,
+            left_phone=left_phone,
+            right_phone=right_phone,
+        )
+        >= 2
+    )
+
+
+class _UnionFind:
```
ColinToft, line 131: Cool!! Learned something new today

backend/python/app/services/implementations/location_import_validation.py
```diff
@@ -0,0 +1,192 @@
...
+    return name.strip().casefold()
+
+
+def _address_match_key(address: str | None) -> str | None:
+    if not present_str(address):
+        return None
+    return address.strip().casefold()
+
+
+def _phone_match_key(phone: str | None, normalized_phone: str | None) -> str | None:
```
ColinToft: I'm a little bit confused here as to why we are taking in a phone number twice? And similarly in `rows_are_duplicates`, what is the reasoning to taking in the phone number separately?

(This isn't really a review comment so much as a clarification question, I'm assuming there's a reason for this)

backend/python/app/services/implementations/location_service.py
```diff
@@ -449,79 +454,99 @@ async def review_locations(
...
-                            location.phone_secondary
-                        )
-                except ValueError:
-                    alerts.append(AlertCode.INVALID_FORMAT)
-                    rows.append(
-                        LocationImportRow(row=row_num, location=location, alerts=alerts)
+                parsed_rows.append((row_num, self._parse_row(row, column_map)))
+
+            geocode_cache: dict[str, GeocodeResult | None] = {}
+            unique_addresses = {
```
ColinToft, line 502: Another clarification, how does this unique address geocoding work? It kinda looks like this is deduplicating within the spreadsheet (which is great) but it would also be fantastic if we could avoid geocoding something that is already in the DB - if it doesn't already do this would it be possible to have that (if it's not to much effort)? The main use case will be importing a spreadsheet where most of the addresses already exactly match something that is in the DB.

ColinToft (approved):
Great work! Left a couple comments but otherwise awesome to see this get done and thanks for making the frontend changes too!!

backend/python/app/services/implementations/location_import_validation.py
Kerushani: Good catch - that is confusing. `left_phone` / `right_phone` weren’t a second phone field; they were meant to be the precomputed E.164 form of phone_primary. But since `review_locations` already normalizes and writes that value back onto the entry before duplicate detection, the separate args were redundant. I’ve removed them so matching just uses `left.phone_primary` / `right.phone_primary` on the import entries.

backend/python/app/services/implementations/location_service.py
Kerushani: Good point. Previously it only deduped within the spreadsheet (unique address strings → one Google call each). It didn’t check the DB.

I added that. Before geocoding, we look up exact matches in `locations.address` that already have lat/lng, treat those as valid, and only call Google for the rest.

One caveat: on ingest/create we store Google’s `formatted_address`, not the raw spreadsheet cell. So this helps when the sheet already matches what’s in the DB (your main use case), but a raw "123 Main St" won’t skip geocoding if the DB has "123 Main Street, Waterloo, ON...". If that becomes an issue we can later match on a normalized key.

frontend/src/pages/admin/routes/generation/ValidateStep.tsx
Kerushani: Good point - will use the errors instead.

frontend/src/pages/admin/routes/generation/ValidateStep.tsx
Kerushani: Yes, I see the existing code did do that - but I'll make the change so multiple alerts appears.

PR #189, opened 2026-07-10
Guard delivery-type removal + add rename flow

ColinToft:
Stacked on `f4krp-199-delivery-types-settings`.

## Why
That branch made delivery types configurable data (a `list[str]` on `SystemSettings`), but validation was **one-directional**: new locations are validated against the list, yet nothing stopped an admin from *removing* a type that locations still use. A removed-but-in-use type orphans those location rows — they become filter-invisible and can no longer round-trip through the edit/import paths (both re-validate against the configured list).

## What
- **Removal guard** — `PATCH /system-settings` now blocks dropping a delivery type that any **on-roster** (Active/Unscheduled) location still uses → `409`. Off-roster (Inactive) locations keep their historical string and don't block removal — their value is historical, and re-import would reassign a configured type anyway. The guard also runs when no settings row exists yet (compares against the defaults), since locations can already reference those.
- **Rename flow** — `POST /system-settings/delivery-types/rename` renames a type in place, cascading the new name onto **every** location using it (active *and* inactive, since a rename preserves identity) and swapping the name in the list. A plain list-PATCH that swaps a name reads as remove+add and is blocked by the guard, so renames must come through this endpoint. Rejects unknown source, blank target, and renaming onto an existing type (which would silently merge two types).

## Design notes
- `in_roster` (the stored bit behind the derived Active/Unscheduled/Inactive status) is the guard threshold — it's exactly the set of locations that would break, and it's a cheap stored column (no future-route query needed).
- Backend-only: the settings-management UI doesn't exist yet (the admin settings page is still a stub), so there's no frontend consumer to wire — just the regenerated OpenAPI client so the contract is captured. UI belongs with the settings-management ticket.

## Tests
`TestSystemSettingsRoutes` now covers: removal blocked when an active location uses the type; removal allowed when unused; removal allowed when only an inactive location uses it; rename cascade to active and inactive locations; rename rejection for unknown source and existing target. Full backend suite green (ruff + `ruff format --check` + mypy + 127 route tests); frontend lint + tsc clean.


ColinToft:
Reopened as #192 — this PR was auto-closed when its stacked base branch `f4krp-199-delivery-types-settings` was deleted after merging into `main`. GitHub won't let a closed PR's base be retargeted, so #192 carries the same change rebased onto `main`.

PR #195, opened 2026-07-10
Rework fleet routing objective; clamp oversized locations to capacity

ColinToft:
Three changes to the `optimizeTours` payload in `GoogleMapsFleetRoutingAlgorithm` and makes a couple small changes. I did use claude to make this oops 😳 

### 1. Objective: minimize when the last driver finishes

Sets cost to:
- **`globalDurationCostPerHour = 6`**. biggest penalty here, basically this is penalizing the longest route
- **`costPerHour = 1`** on every vehicle: a smaller cost on the driving time of each vehicle

Really only the ratio matters: at 6:1 the optimizer spends up to 6 extra total driver-hours to get the last driver home an hour sooner, but we can maybe tweak this.

### 2. Use actual boxes (not children) as the unit

Now that the child per box is in settings, we can use that

### 3. Clamp oversized locations to vehicle capacity

A location whose box count exceeds vehicle capacity (e.g. an 18-box school with 12-box cars) previously broke things, I solved this with the big brain strategy of just round down to the capacity (in this case 18 rounds down to 12).

## Testing

- Unit tests: objective fields on every vehicle + model; box derivation incl. ceil rounding and custom `children_per_box`; clamp at/below/above capacity boundaries (16 pass)
- `ruff check`, `ruff format --check`, `mypy` all clean
- OpenAPI schema + generated frontend client regenerated


backend/python/tests/test_google_maps_routing_service.py
```diff
@@ -91,8 +93,12 @@ def test_basic_payload_structure(
             assert v["displayName"] == f"driver_{i}"
             assert v["startLocation"] == {"latitude": 43.0, "longitude": -79.0}
             assert v["loadLimits"] == {
-                "load": {"maxLoad": str(sample_settings.max_half_boxes_per_driver)}
+                "load": {"maxLoad": str(sample_settings.max_boxes_per_driver)}
             }
+            assert v["costPerHour"] == VEHICLE_COST_PER_HOUR
+
+        # --- objective: makespan pressure on the whole plan ---
```
kingMonkeh: nit: what does this comment even mean 💀 
or i guess what is a makespan??

backend/python/tests/test_google_maps_routing_service.py
```diff
@@ -171,38 +177,75 @@ def test_no_return_to_warehouse_omits_end_location(
         vehicle = payload["model"]["vehicles"][0]
         assert "endLocation" not in vehicle
 
-    def test_route_duration_limit(
+    def test_makespan_dominates_per_vehicle_cost(self) -> None:
```
kingMonkeh: forgot to flag this one as well 😭

kingMonkeh (approved):
LGTM 🤑

backend/python/tests/test_google_maps_routing_service.py
ColinToft: oops fixed

PR #196, opened 2026-07-11
Remake driver history as an append-only mileage ledger

ColinToft:
## Summary

`DriverHistory` becomes an **append-only ledger** — one row per mileage event (`AUTO` delivery credit, `REASSIGNMENT` compensation, `MANUAL_ADJUSTMENT`) instead of an opaque monthly running total. Monthly/lifetime totals are `SUM(km)` aggregates bucketed by `drive_date`. Every km is traceable to the delivery or correction that produced it; corrections append, never overwrite.

Spec discussion: this dissolves the full edge-case map we audited (late assignment, reassignment after freeze, driverless completions, deletions, missed nightly runs, manual-edit conflicts).

## What changed

**Ledger + reads**
- `driver_history` table: `driver_id` (SET NULL), `route_id` (SET NULL), `drive_date`, signed `km`, `kind`, `note`. Partial unique index: at most one AUTO credit per route (structural no-double-count guard).
- Monthly list / summary / by-year / CSV export become aggregates; **API response shapes unchanged**. New `GET /drivers/{id}/history/entries` audit view.
- Manual `PATCH`/`DELETE`/`POST` (hard-set) endpoints replaced by admin-only `POST .../adjustments` (signed delta + required note). Drivers can no longer edit their own km (was `SELF_DRIVER_OR_ADMIN`).

**Nightly freeze job**
- Scans **all due un-frozen routes** (`drive_date <= today`), not just today's → missed runs, crashes, and unconfigured warehouse coords self-heal on the next run.
- One transaction per route: snapshot + stop snapshots + AUTO credit land together or not at all. One failing route can't poison the run.
- `RouteSnapshot.length_km` records the credited-at-freeze length.

**Reconciliation on frozen-route edits ("correct the record")**
- Driver change → compensating `REASSIGNMENT` entries moving exactly the credited `snapshot.length_km` (null→B backfills, A→B moves, A→null reverses).
- Stop edits → stop snapshots **rebuilt** (previously cascade-destroyed), `length_km` updated, km delta posted to the current driver. Combined patches net the new driver exactly the corrected length.

**Deletion policy**
- Route/group delete keeps mileage (`route_id` → NULL): the delivery happened.
- Driver delete is now a **soft-delete** (`active=false`) so history keeps the name; ledger FK SET NULL as backstop.

**Guards**
- `drive_date` locked once a group has frozen routes (409).
- Deleting a location referenced by route stops → clean 409 listing the routes (was raw FK 500).

**Bonus fix**: CSV export previously showed each driver's *last month's* km as their yearly total (dict-comprehension overwrite); now a real yearly sum.

## Migration

Backfills AUTO entries from frozen driven routes, then preserves **every legacy (driver, year, month) total exactly** via per-bucket reconciliation adjustments (FULL OUTER JOIN — legacy is source of truth since admins may have manually corrected it). Validated **up → down → up** against a real Postgres 17 with a manual-drift case: totals identical at every step. `route_snapshots.length_km` backfilled from live route lengths.

## Testing

- 25 new tests: ledger arithmetic, full reassignment matrix (frozen + unfrozen), frozen-edit amendments (incl. combined reassign+amend), catch-up freeze, per-route atomicity (real failure injection via the unique index), warehouse self-heal, deletion survival, soft-delete, both guards, adjustment API validation.
- Existing tests updated where they encoded the old contract (monthly CRUD, hard driver delete, snapshot fixtures, auth matrix, seed schema).
- `ruff check` + `ruff format --check` + `mypy`: clean. Full suite: **zero regressions vs main** (the 68 pre-existing env-dependent failures on main are unchanged).
- OpenAPI client regenerated.


ColinToft:
Closing in favor of a simpler derived-mileage design: rather than maintaining a ledger, mileage will be computed directly from frozen routes (SUM(route.length) grouped by driver/month) plus a small adjustments table for manual corrections and pre-app history. Routes are kept forever (storage is trivial without deleting them), which dissolves the ledger's main justification. Replacement PR incoming.

PR #201, opened 2026-07-14
Lock down driver self updates

aaronwxng:
## Summary
- Let drivers update their own first name, last name, and phone through the existing driver update endpoint
- Block self-driver updates to admin-only fields like address, availability, vehicle details, and active
- Keep admin updates unrestricted and update generated API types

## Tests
- docker compose exec backend ruff format app tests scripts
- docker compose exec backend python -m pytest tests/test_routes.py tests/test_auth_middleware.py::TestRequireSelfDriverOrAdmin tests/test_models.py -q
- corepack pnpm@9.15.9 lint
- corepack pnpm@9.15.9 build

### DIFF
--- a/backend/python/app/services/implementations/driver_service.py
+++ b/backend/python/app/services/implementations/driver_service.py
@@ -1,7 +1,8 @@
 import logging
-from typing import Any
+from typing import Any, ClassVar
 from uuid import UUID
 
+import firebase_admin.auth
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import selectinload
 from sqlmodel import select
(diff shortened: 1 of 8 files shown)


backend/python/app/services/implementations/driver_service.py
```diff
@@ -148,10 +148,17 @@ async def update_driver_by_id(
                 return None
 
             update_data = driver_data.model_dump(exclude_unset=True)
-            old_values = {field: getattr(driver, field) for field in update_data}
+            user_update_fields = {"first_name", "last_name"}
```
ColinToft: This set is hardcoded multiple times, it would be nice (if possible) to define this list in one place

ColinToft (approved):
LGTM!!

PR #207, opened 2026-07-19
F4KRP-249: Fix Announcement Last Read to use DB column

Kerushani:
## JIRA Ticket
[F4KRP-249
](https://f4kblueprint.atlassian.net/browse/F4KRP-249)
## Implementation Description
Hooks the announcements board’s New badge up to the existing db column (announcement_last_reads.last_read_at) from #199, instead of tracking reads in localStorage.
- Open board → GET /announcements/ returns is_read per post (created_at <= last_read_at)
- New badge → unread (is_read !== true) and posted within 7 days
- Close board → POST /announcements/mark-read advances last_read_at so those badges clear next time
- 7-day expiry → badge disappears even if never “seen” once older than 7 days

### `frontend/src/types/announcement.ts`
- **Added:** optional `is_read?: boolean | null` on `Announcement`

### `frontend/src/api/announcements.ts`
- **Added:** `useMarkAnnouncementsAsRead()` wrapping `markAnnouncementsAsReadMutation`
  - Optimistic update: set all cached announcements to `is_read: true`
  - Rollback on error; invalidate list on settle

### `frontend/src/features/announcements/useAnnouncementReads.ts`
- **Deleted:** localStorage helpers (`STORAGE_PREFIX`, `storageKey`, `loadReadIds`, `persistReadIds`), `readIds` state, per-id `markAsRead(userId)`
- **Added:** `markBoardAsRead()` → `POST /announcements/mark-read`

### `frontend/src/features/announcements/utils.ts`
- **Changed:** `isAnnouncementNew(announcement)` - no longer takes `readIds`
- **Deleted:** `readIds.has(announcement_id)` check
- **Added:** `announcement.is_read === true` → not New; still requires within 7 days

### `frontend/src/features/announcements/AnnouncementCard.tsx`
- **Deleted:** `readIds`, `onOpen` props; subject/message no longer buttons
- **Changed:** New badge via `isAnnouncementNew(announcement)` only
- **Why:** reads aren't per-click anymore (as per the Figma design; only user editable announcements should be noticeably hoverable); closing the board marks read. Edit/delete menu unchanged

### `frontend/src/features/announcements/AnnouncementsPanel.tsx`
- **Deleted:** `readIds`, `onAnnouncementOpen` props (and passthrough to cards)
- **Why:** panel only closes; board owns mark-read

### `frontend/src/features/announcements/AnnouncementsBoard.tsx`
- **Changed:** `currentUserId` from auth store (was `''`)
- **Deleted:** `readIds` / `handleAnnouncementOpen` / passing those to the panel
- **Added:** `closePanel()` → close UI + `markBoardAsRead()`

## Steps to Test

1. **Fresh / unread**
   - Create or use announcements from the last 7 days
   - Open the board → those should show **New** (`is_read: false`)

2. **Close clears New**
   - Close the panel (X, backdrop, or Escape)
   - Reopen → previously New items should not show New (mark-read ran; `last_read_at` updated)

3. **New after watermark**
   - With board closed, create another announcement
   - Open board → only the new one shows **New**

4. **7-day expiry**
   - An announcement older than 7 days with `is_read: false` should not show **New**

* Is the functionality make sense? Does the new badge appear and disappear as it should?

## Checklist

- [x] Follows Tailwind/shadcn best practices per README; no hardcoded color or spacing values
- [x] Implementation matches Zeplin designs and screenshots are attached above
- [x] Screens render correctly on mobile and desktop; no console warnings or errors


Kerushani:
Claude: _Mark-read fires unconditionally on every panel close, including instant closes (e.g. Escape immediately after opening, before isLoading resolves) or closes triggered by clicking the backdrop. That means announcements can be marked read without the user actually having seen them rendered. This matches the PR description's stated design ("Close board → advances last_read_at"), so it's likely intentional, but worth double-checking it's the desired UX versus, say, only marking read after the panel has been open for a moment._

I'm leaning towards keeping the _Treat  “opened the board”_ as enough, even for a quick Escape.

Pros: Simple
Cons: Can clear New without the list ever rendering

Alternatives would be:
- Minimum open duration: only mark-read if the panel was open ≥ 500ms–1s (and ideally after load). Cons: Arbitrary threshold; slow readers vs fast skimmers both get edge cases

frontend/src/features/announcements/useAnnouncementReads.ts
```diff
@@ -1,46 +1,14 @@
...
-    [userId]
-  );
+export function useAnnouncementReads() {
+  const { mutate, isPending } = useMarkAnnouncementsAsRead();
 
-  const readIdSet = useMemo(() => readIds, [readIds]);
+  const markBoardAsRead = useCallback(() => {
+    if (isPending) return;
+    mutate({});
+  }, [mutate, isPending]);
```
ColinToft: ok was just talking to claude lol. i think we can delete this file, and then in announcementboard.tsx:

```
// AnnouncementsBoard.tsx
const { mutate: markBoardAsRead } = useMarkAnnouncementsAsRead();

const closePanel = () => {
  setPanelOpen(false);
  markBoardAsRead({});
};
```

which just uses the tanstack thing directly

frontend/src/api/announcements.ts
```diff
@@ -55,3 +57,27 @@ export function useSendAnnouncementEmail() {
     ...sendAnnouncementEmailMutation(),
   });
 }
+
+export function useMarkAnnouncementsAsRead() {
```
ColinToft, line 60: also pretty unsure about this tbh but claude says we can:

```
export function useMarkAnnouncementsAsRead() {
  const queryClient = useQueryClient();
  return useMutation({
    ...markAnnouncementsAsReadMutation(),
    onSuccess: () => {
      queryClient.invalidateQueries({ queryKey: getAnnouncementsQueryKey() });
    },
  });
}
```

which apparently loses the unread badge being cleared instantly but that is prob ok because my understanding is we mark as read when we close the board (so we don't see this stuff anyway)

ColinToft (approved):
LGTM! a couple comments that are mostly from a helpful AI assistant lol

frontend/src/api/announcements.ts
Kerushani: I agree - this simplifies the code. I can add that in.

frontend/src/features/announcements/useAnnouncementReads.ts
Kerushani: Makes sense - I deleted useAnnouncementReads and I'm using the tanstack thing instead

PR #206, opened 2026-07-18
Scope import stale handling by delivery type

aaronwxng:
## Summary
- Adds delivery type to the location import review request so stale/changed matching is scoped to the selected type.
- Adds the delivery type selector to the Import step and only shows file upload after a type is selected.
- Sends the selected delivery type through review and ingest, and uses the real review results when applying import changes.

## Tests
- docker exec f4k_backend ruff check .
- docker exec -e PYTHONPATH=/app f4k_backend mypy .
- docker exec -e PYTHONPATH=/app f4k_backend pytest tests/test_routes.py::TestLocationImportRoutes -q
- docker exec -e PYTHONPATH=/app f4k_backend pytest tests/test_routes.py -q
- eslint . --max-warnings=0
- tsc -b
- vite build

PR #209, opened 2026-07-21
Build data import changes review page

aaronwxng:
## Summary
- Shows the real net-new, removed, and changed rows from the import review response.
- Lets admins choose what to do with each row before committing the import.
- Applies changes through the existing ingest endpoint and sends changed rows marked as separate as old stale + new net-new rows.
- Allows CSV uploads in the Import step to match the PRD.

## Tests
- eslint . --max-warnings=0
- tsc -b
- vite build
- git diff --check

Note: this PR is stacked on #206 and targets its branch so the diff only shows this frontend ticket.


ColinToft:
@claude review

PR #210, opened 2026-07-21
[DO NOT MERGE] Changes for Backend & Frontend Deployment

ludavidca:
## JIRA Ticket

## Implementation Description

*

## Steps to Test

1.

## Screenshots *(frontend only — delete if not applicable)*

| Implemented | Zeplin Design |
|-------------|---------------|
|             |               |

*

## Checklist


ludavidca:
@claude review

ColinToft:
@claude review

ColinToft:
Claude comment haha but keeping it here cause it has some useful things.

Went through this properly — thanks for pushing the deployment story forward, this is the part of the project nobody else has touched and a lot of it is right.

Some context that should make it easier rather than harder: I'm putting up a PR that unifies the environment config, which removes most of what you're building on top of here.

**Why that matters for this PR.** `ENVIRONMENT` is never actually set anywhere in the repo — only `APP_ENV` is. So `settings.environment` always falls back to `"development"`, which means `settings.is_production` is *always* False and `settings.is_development` is *always* True, in every environment. On a real deploy that means the refresh cookie ships without `Secure` and `/docs` + `/redoc` are publicly reachable. Not something you introduced — it's been latent since the starter code, and this PR is just the first thing that would expose it.

The fix collapses `Settings` down to one class with an `Environment` enum keyed off `APP_ENV`, and deletes `DevelopmentSettings` / `ProductionSettings` / `TestingSettings` (they only existed to set `debug` and `testing`, which nothing in the backend reads). That should simplify the secret-mount code here a lot — no `target_class`, no `__fields__` filtering, just load the JSON into the one `Settings`. It also fixes this, which I think is a real bug as written:

```python
target_class = DevelopmentSettings
if environment == "production":
    target_class = ProductionSettings
...
allowed_fields = target_class.__fields__.keys()
...
return Settings(**cloud_secrets)   # target_class computed, then discarded
```

Happy to do that rebase for you once mine lands if it's easier — just say the word.

---

Separately, a few things in here would land on everyone's local setup, and I think they're worth pulling out before this merges:

**1. `pull-env.sh` stops working.** The new comment sits inside a `\` line continuation:

```bash
if gcloud secrets versions access latest \
# replace with food4kids-config-dev if pulling dev deployment secrets
  --secret="f4k-development-backend-env" \
```

Bash splices continuations *before* stripping comments, so this runs `gcloud secrets versions access latest` with no `--secret`, and then tries to execute `--secret=... --project=... > .env.tmp` as its own command. I reproduced it — exit 127 and an empty `.env.tmp`. Moving the comment above the `if` fixes it.

**2. `?sslmode=require` on the local dev path** (`app/models/__init__.py`). The comment above it says "safe for localhost, no forced SSL" and then the `else` branch appends it unconditionally. Local Postgres containers don't have SSL configured, so this breaks `asyncpg` for anyone running the normal docker setup. I think the `is_testing` / `else` branches may have got swapped — and note `is_testing` is always False today for the reason above, so the test DB path currently can't be reached at all.

**3. `server.py` `reload=True` → `reload=False`** turns off hot reload for local development. If Cloud Run needed this, could it go in the `else` (production) branch instead?

**4. `frontend/package-lock.json`** (17k lines) — the repo is on pnpm, and #253's workflow does `npm ci` against this. Two lockfiles will drift silently. Could the deploy workflow use pnpm instead?

**Also worth stripping before this goes out:**

- `print(f"--- DIAGNOSTIC PWD_LEN: {len(pwd)} | PWD_START: '{pwd[:3]}' ---")` in `config.py`, plus the `--- DEBUG:` prints of host/db/user in `migrations/env.py` — Cloud Logging keeps those and they're fairly broadly readable.
- The `except Exception` around the secret parse warns and then falls through to `return Settings(**cloud_secrets)` with `cloud_secrets == {}`, so a malformed secrets file boots the app with empty DB credentials. I'd rather it crash on startup — a container that won't start is much easier to diagnose than one that's up and silently misconfigured.
- `migrations/env.py` now reimplements the secrets-file read a second time, and the two copies already disagree (it reads `DATABASE_URL` off the raw uppercase dict, `config.py` lowercases first). One shared helper would keep them from drifting.
- `temporary-frontend/dist/` build artifacts are committed.

**On the `/api` prefix** — the change itself looks right to me, and the Firebase rewrite is a good call for keeping everything same-origin. It's just the highest-conflict thing in flight: it touches every router, `openapi.json`, the generated SDK, and the conftest base URLs. There are six other PRs open right now. If we land it, I'd want it to go in on its own and first, with everything else rebasing onto it, rather than bundled with the secret loading and the Firebase config.

Which is basically my one structural suggestion: this is four separable changes (env/secret loading, the `/api` prefix, Firebase hosting config, and the local-dev fixes). Split into four they can each land as they're ready — right now the whole thing is blocked on whichever piece is least finished, and it's picked up merge conflicts sitting there.

Let me know what would be most useful — happy to take any of these on myself.

ColinToft:
Closing because this is superseded by #280-283 and #315

PR #214, opened 2026-07-25
Debug: surface Claude review error output (temporary)

ColinToft:
Temporary diagnostic — **do not merge**.

Claude Code Review has failed on every run since 2026-07-21 (~2s, `$0` cost, no assistant turn, `is_error: true`). `claude-code-action` logs only a summary of the SDK result message, so the actual error string is invisible. This adds an `if: always()` step that prints the result object from the execution file.

Opening this PR triggers the head-branch workflow, which should reveal the real error. Will close once we have it.


ColinToft:
Closing — the action refuses to run when the workflow file differs from main (deliberate guard against exfiltrating secrets via a PR that edits the workflow), so this diagnostic can't work from a branch. Cause is being tracked down another way.

PR #219, opened 2026-07-26
Wilson/routes all screen groups routes & addresses tabs

wilson-li3:
## JIRA Ticket

## Implementation Description

*

basically all the tabs in routes. routes, groups, and addresses. all functionality was implemented. only thing missing is filter for routes

## Steps to Test

just make sure everything functions and properly updates when things are edited, added, deleted, stuff like that

*

## Checklist


frontend/src/pages/admin/routes/components/AddRouteGroupModal.tsx
```diff
@@ -0,0 +1,174 @@
+import { useState } from 'react';
```
ColinToft: I think we can delete this file since we aren't going to support adding a route group

backend/python/app/services/implementations/route_service.py
```diff
@@ -104,18 +124,67 @@ async def get_routes(
...
+        )
+        has_stops_subq = (
+            select(1)
+            .select_from(RouteStop)
+            .where(col(RouteStop.route_id) == col(Route.route_id))
+            .correlate(Route)
+        )
+        # NOTE: delivery types are configurable (SystemSettings.delivery_types)
+        # since #181; "School"/"Family" here match the seeded defaults only.
+        delivery_type_expr = case(
```
ColinToft: Can we fix this? I think remove has_school_subq and stuff. We can assume that a route has all of the same location type in it (an invariant) so no need to prioritize school over family. Essentially we just want to support any delivery type name (so "School" or "Family" should appear nowhere in this PR)

frontend/src/pages/admin/routes/components/RouteGroupsTab.tsx
```diff
@@ -62,114 +83,114 @@ export function RouteGroupsTab({
...
-        </div>
-        <div className="flex items-center gap-4">
+      <TableToolbar
+        search={search}
+        showFilter
+        onFilterClick={openFilters}
+        hasActiveFilters={hasActiveFilters}
+        actions={
           <Button variant="primary" asChild>
             <Link to="/admin/routes/generation">Generate Routes</Link>
```
ColinToft: Image

A few casing things (isabelle said buttons should be sentence case, we recently updated designs)

ColinToft (approved):
Good work! A couple things that didn't really have a specific place in the code:

No pagination: All three design tabs show 1 2 3 … 10 Next at the bottom (can you add that)

Tab state isn't in the URL right now. AdminRoutesPage.tsx uses defaultValue="groups" with no sync. Refresh, deep-link, and browser Back all land on Groups. This will be needed I think since dashboard/other places may want to link to a specific tab

(Also it currently is not passing CI)

I'm ok with you merging it once these are fixed!

PR #220, opened 2026-07-27
Implement Sentry traces and logging for the frontend

ChloeW125:
## JIRA Ticket
Implement Frontend Logging via Sentry (for Deployment)(https://f4kblueprint.atlassian.net/browse/F4KRP-253)

## Implementation Description
* Implement Sentry tracing and logging to track errors, traces, and logs for the frontend

## Steps to Test
1. If you run the frontend and perform any action that produces logs (info, error, warning logs, etc.), you should see relevant log entries in Sentry (will need F4K Admin account) 
2. If you run the frontend and perform any type of action, you may see relevant traces in Sentry (will need F4K Admin account) - note that Sentry only traces about 20% of user flows (to make sure we can stay under the Free Tier budget) so if you don't see anything in Sentry traces after interacting with the frontend, that may be why
3. Check package.json and pnpm-lock.yaml only has necessary changes made to it

* Frontend still runs!
* Logs appear in Sentry (when applicable)

## Checklist


ChloeW125 (commented):
@claude review

ludavidca (approved):
LGTM

PR #208, opened 2026-07-21
Implement Create / Forgot Password

kingMonkeh:
## JIRA Ticket
[F4KRP-183](https://f4kblueprint.atlassian.net/browse/F4KRP-183)

## Implementation Description

* Built Create Password Screens
* Built Forgot Password Screens
* Created WrapperWithLogo component to stay DRY
* Create PasswordResetToken table
* Create PasswordResetToken service + routes
* Write frontend auth hooks for create / forget password
* Updates tests

## Steps to Test

1. Begin by initializing a driver in the backend using Postman
2. You should receive an email to create your account, follow it
3. After creating your account, try logging in with your credentials, it should work
4. Go back to the login screen and click "Forgot password"
5. Enter your email
6. You should receive an email containing a password reset link
7. Follow it and reset your password
8. Once again, try to login using your new password

## Screenshots *(frontend only — delete if not applicable)*

| Implemented | Zeplin Design |
|-------------|---------------|
|             |               |

All the screens circled are implemented
image

## What Should Reviewers Focus On?

* I started this thing, where like should you be able to see the page even if your token is not valid? 
I implemented 2 different versions, one that allows invalid tokens (create password)
And another that validates token on page load (Reset password) 
I'd like an opinion on which we should follow, I think validating on page load is lowk optimal despite requiring a bit more work

## Checklist
- [ ] PR name and commits are descriptive, imperative, and atomic (trivial commits squashed)
- [ ] I have requested a review from Claude, understood its suggestions, and implemented fixes where needed (or documented disagreements)
- [ ] I have requested a review from the PL and relevant devs with background on this PR

**Frontend only — delete if not applicable:**
- [ ] Follows Tailwind/shadcn best practices per README; no hardcoded color or spacing values
- [ ] Implementation matches Zeplin designs and screenshots are attached above
- [ ] Screens render correctly on mobile and desktop; no console warnings or errors

PR #211, opened 2026-07-23
F4KRP-202 Created driver individual routes page

eddywang4340:
## JIRA Ticket
[Individual Driver Routes Page](https://f4kblueprint.atlassian.net/browse/F4KRP-202)

## Implementation Description

- Frontend
  - New driver Individual Route page (/driver/route) - matches Figma
  - Route header (name, date · start time, meta row: stops / type / distance / boxes), map view, and stops list
  - PDF (browser print), Call Food4Kids (tel: placeholder number, TODO), Google Maps (backend directions link) buttons
  - Stops render as collapsible cards (street on top, city · boxes below); expanded body is a TODO: Add Notes Rendering placeholder (notes/images out of scope)
  - Temporary dev-only "Route ID (dev)" input to load a route (marked with TODO to remove once driver nav lands)

- Backend
  - GET /routes/{id} now returns drive_date and delivery_type on RouteDetailRead
  - Regenerated OpenAPI client + tests updated

## Steps to Test

- Checkout this branch
- Run the frontend
- Make sure you have a route in your DB (if not, use the seeding DB script). Grab its route_id, and make sure the route has an encoded polyline. If not, use Swagger UI to hit the routes PATCH endpoint and it'll generate one for you
  - NOTE: Don't hit this endpoint too much it uses the Google Maps API key (cost involved!)
  - TIP: To get the location IDs for a route (needed by the PATCH endpoint):
docker-compose exec db psql -U postgres -d f4k -c "SELECT location_id FROM route_stops WHERE route_id = '<your-route-id>' ORDER BY stop_number;"
- Go to /driver/route, paste the route_id into the "Route ID (dev)" input, and click Load
- You should see the route detail page: header (name, date · start time, stops/type/distance/boxes), map view, and the list of stops
- Verify the actions: PDF opens the print dialog, Call Food4Kids is a tel: link, Google Maps opens directions in a new tab, and expanding a stop shows the TODO: Add Notes Rendering placeholder

## Screenshots *(frontend only — delete if not applicable)*

- Implemented (desktop)
image

- Figma (desktop)
image

- Implemented (mobile)
image

- Figma (mobile)
image

## What Should Reviewers Focus On?

*

## Checklist
- [x] PR name and commits are descriptive, imperative, and atomic (trivial commits squashed)
- [x] I have requested a review from Claude, understood its suggestions, and implemented fixes where needed (or documented disagreements)
- [ ] I have requested a review from the PL and relevant devs with background on this PR

[...]

PR #222, opened 2026-07-28
Make driver delete actually delete the driver

ColinToft:
## The bug

`DELETE /drivers/{driver_id}` removed only the `drivers` row. The `users` row, the `user_invites` row and the Firebase account all survived — so a "deleted" driver could still `POST /auth/login` and get back a valid access token, and still resolve by email for a password reset. The PRD calls for a hard delete, so this was a divergence, not a deliberate soft delete.

Pre-existing; **not** introduced by #208.

## The fix

`delete_driver` is now the inverse of `initialize_driver`: it deletes the `users` row and lets the existing FK cascades take `drivers`, `user_invites`, `announcement_last_reads` and `announcements`, reusing `UserService.delete_user_by_id` for the Firebase half.

**Ordering** — `delete_user_by_id` already had it right: delete the Firebase credential, *then* commit, in one DB transaction. A Firebase failure rolls the DB work back into a retryable no-op. The other order would leave a live login for a driver the admin has been told is gone, which is the failure that actually matters. `initialize_driver` makes the same trade in the other direction (DB rows commit, then the invite email dispatches, so a send failure never leaves a half-built driver).

That method had **no callers** before this, so its async cascade path was untested — `User.driver` now loads eagerly, because a lazy load during an async flush raises `MissingGreenlet`.

## Two things enumerating the FKs uncovered

| FK to `users` | before | after |
|---|---|---|
| `drivers.user_id` | CASCADE | — |
| `user_invites.user_id` | CASCADE | — |
| `admin_info.user_id` | CASCADE | — |
| `announcements.user_id` | CASCADE | — |
| `announcement_last_reads.user_id` | CASCADE | — |
| `notes.user_id` | **NO ACTION** | **SET NULL** |

- **`notes.user_id` had no ON DELETE action.** Deleting a user who had ever written a note failed with a FK violation — the delete was *impossible* for any driver who had actually done the job. Migration `c9a1e5f30b74` makes it `SET NULL`. A note is an operational record ("gate code is 4821") that outlives its author; the column was already nullable and readers already render an authorless note, so this is both expressible and the privacy-preserving choice for a hard delete.
- **The driver's admin-only note chain would be orphaned.** Every driver gets one at initialize, and `drivers.note_chain_id` is `SET NULL`, so the chain would survive with nothing referencing it — notes *about* a deleted person that no endpoint could ever reach or delete. It now goes in the same transaction, via a `delete_note_chain_rows` split out of `NoteChainService`.

Routes stay detached rather than deleted (unchanged): the delivery record survives, and derived mileage goes to zero on its own.

Also drops `DriverService.delete_driver_by_id` and its unused sibling `delete_driver_by_email` — both implemented the same incomplete delete.

## Tests

New `tests/test_driver_delete.py` (12 tests). **Eight fail against the old handler** — verified by stashing the fix and re-running.

- user + invite + Firebase account all removed
- a deleted driver cannot log in — the sign-in call is mocked to *succeed*, so this asserts the DB half independently of the Firebase delete
- no `users` row survives to resolve a password reset by email
- driver who never completed signup (`auth_id IS NULL`): 204, and Firebase is not called
- non-existent driver → 404; deleting twice → 204 then 404
- Firebase failure rolls the whole thing back, and the retry then succeeds
- the driver's note chain and its notes go; notes they wrote elsewhere survive with `user_id` NULL
- routes detach rather than delete; delete is admin-only

## Verification

- `pytest tests/` — **509 passed** (497 baseline + 12 new)
- `ruff check` ✅ · `ruff format --check` ✅ · `mypy` ✅ · frontend `eslint src` ✅
- `alembic upgrade head` from an empty DB ✅, and `downgrade -1` round-trips ✅
- `frontend/openapi.json` + generated TS regenerated

Run against this branch's code, not the shared `f4k_backend` container — that one bind-mounts the main folder, which is currently on `create-forget-password`.

## One thing to decide separately

Should `/auth/login` (and `/auth/forgot-password`, once #208 lands) refuse a user row that has neither a driver nor an admin role? That is defence in depth against this whole class of orphan, and it is cheap. **Deliberately not bundled here** — it is a policy change to the auth surface, not part of making delete correct. Happy to open it as its own PR if you want it.

Note on the original report: `/auth/forgot-password` does not exist on `main` — it arrives with #208 (`create-forget-password`), which is what the local stack was running. On `main` the equivalent surface is `/auth/resetPassword/{email}`, which is authenticated. The `users`-row mechanism behind step 5 is the same either way, and is what this PR removes. #208's `password_reset_tokens.user_id` is already `ON DELETE CASCADE`, so it composes with this fix without changes.


ColinToft:
Rebased onto main now that #208, #211, #219 and #221 have landed. Two CI failures on the first push, both from being behind main rather than from the fix:

- **`boot`** — two alembic heads. My migration parented onto `b3d4a1c2e5f6`, which main's new migrations moved past. Re-parented onto `bc7876e56dd1`; single head again, verified `upgrade head` from an empty DB and `downgrade -1`.
- **`test`** — `AuthService.send_create_password_email` is gone; `initialize_driver` dispatches through `EmailDispatcher` now. Test helper updated.

#208 landing also made the reported step 5 **directly** testable rather than by proxy. `test_deleted_driver_cannot_obtain_a_password_reset_token` now drives the real `/auth/forgot-password`. It answers 204 either way by design (anti-enumeration), so the assertions are that no token row is minted and no email is dispatched — with the Firebase lookup mocked to *succeed*, so the DB check is the only thing under test.

Added `test_reset_tokens_are_deleted_with_the_driver` too. `password_reset_tokens.user_id` was already `ON DELETE CASCADE` so no code change was needed, but a token outliving the account is a live reset link sitting in a deleted driver's inbox — worth pinning.

**522 passed** (509 baseline + 13). Re-ran the vacuity check against main's handler with the updated tests: **9 of 13 fail**, including both reset-token tests.

PR #223, opened 2026-07-28
F4KRP 264: Add changes to the Jobs table

Kerushani:
## JIRA Ticket
[F4KRP 264](https://f4kblueprint.atlassian.net/browse/F4KRP-264) [sub tickets created under parent ticket just to split up some of the logic]

OG "Job Executor" Doc -> [doc](https://www.notion.so/Route-Generation-Job-Executor-Options-and-Recommendation-3a4975f1a25280e4aee1eb6c3418ce1f?source=copy_link) 

Concise and readable "Job-Executor doc":  [Doc](https://www.notion.so/uwblueprintexecs/Route-Generation-Job-Executor-3ab10f3fb1dc80f5b0f6d6cb825e7f48?source=copy_link)

## Implementation Description
* The problem this closes: POST /jobs/generate validates the `RouteGenerationGroupInput` request body via FastAPI, then throws it away, `JobService.generate_job` never used it (the param was even `named _req`). A Job row only ever had progress/timestamp fields, with nowhere to record what to generate, why a run failed, or what it produced.

The plan is to run route generation in the background instead of making the person wait. Calling Google's routing service can take a long time, tens of seconds, and nobody wants to stare at a loading spinner that long. So the plan is: the server says "got it, I'll get started" almost instantly, and a separate background process does the slow work later.

But that background process runs after the original request is already finished and gone. If we don't save the details of what was requested somewhere permanent (like the database), there's nothing left for the background process to work from once the request is over.

What changed, file by file:

`app/models/job.py `-> Added six nullable columns directly on the Job table class (not on the shared JobBase, so JobRead/JobCreate, and the OpenAPI schema/generated TS client, are untouched by this PR):

`app/services/implementations/job_service.py` -> generate_job now stores `req.model_dump(mode="json")` on input_payload when creating the job, instead of discarding req.

`app/routers/job_routes.py` -> Renamed the _req parameter to req since it's actually used now.

## Steps to Test

Run inside the f4k_backend container:
```
# 1. Apply the new migration (adds the new columns to the database)
docker-compose exec backend alembic upgrade head

# 2. Run just the tests related to this change
docker-compose exec backend python -m pytest tests/test_routes.py -k "generate_job" -v
```

* If new columns added to jobs table is valid

## Checklist


ColinToft (approved):
This looks good to me! Thanks for doing this!

PR #235, opened 2026-07-31
F4KRP-255: Configure route groups before generation

aaronwxng:
## Summary
- replace mocked route-group rows with groups created by the completed import
- prefill route dates, start times, route counts, and return behavior from imported data and system settings
- let admins select groups and edit generation settings in the Figma-aligned table
- pass validated per-group generation inputs to the Generate step
- add Figma-aligned confirmation and leave-without-saving dialogs

## Validation
- `docker compose exec frontend pnpm run build`
- `docker compose exec frontend pnpm run lint`
- `git diff --check`

No Postman files were added or changed.

### DIFF
--- a/frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
+++ b/frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
@@ -1,17 +1,14 @@
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 import { Navigate, useNavigate, useOutletContext } from 'react-router-dom';
 
+import { useLocationGroups, useSystemSettings } from '@/api';
 import EditIcon from '@/assets/icons/edit.svg?react';
 import type { Column } from '@/common/components';
 import {
+  Banner,
   Button,
   DataTable,
   DatePicker,
-  Dropdown,
-  DropdownContent,
-  DropdownItem,
-  DropdownTrigger,
-  DropdownValue,
   Input,
   Modal,
   ModalContent,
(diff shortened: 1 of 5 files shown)


frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
```diff
@@ -1,159 +1,350 @@
...
+  friday: 5,
+  saturday: 6,
 };
 
-const END_LOCATION_OPTIONS = [
-  { label: 'Warehouse', value: 'Warehouse' },
-  { label: 'School', value: 'School' },
-] as const;
+function nextRouteDate(groupName: string, today = new Date()): Date {
+  const weekdayName = groupName.trim().split(/\s+/)[0]?.toLowerCase();
```
ColinToft: OK this is kind of interesting! We probably can't assume that all the groups are gonna be called like "Tuesday A", especially at other Food4Kids chapters. but, honestly this could be kinda helpful to set date based on the weekday since it's probably right most of the time. i just asked isabelle about this but i think we should keep it. maybe instead of split() and then [0], we could just find the first weekday occurrence, like find the first match of monday|tuesday|wednesday.... in the group name, just to be a bit more general

frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
```diff
@@ -1,159 +1,350 @@
...
+    const routeDate = nextRouteDate(groupName);
+    return {
+      name: defaultRouteName(groupName, routeDate),
+      routeDate,
+      startTime: normalizeTime(systemSettings?.route_start_time),
+      routeCount:
+        defaultCap && defaultCap > 0
+          ? Math.max(1, Math.ceil(stops / defaultCap))
+          : 1,
+      returnToWarehouse: selectedDeliveryType.toLowerCase().includes('school'),
```
ColinToft: not a huge deal but I think we should probably just default this one to false, don't want to really hard code anything that is relying on delivery types since we want them to be customizable

frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
```diff
@@ -1,159 +1,350 @@
...
   }
 
-  const updateEntry = (key: string, updates: Partial<RouteFormEntry>) => {
-    setFormData((prev) => ({
-      ...prev,
-      [key]: { ...prev[key], ...updates },
-    }));
+  const defaultEntry = (groupName: string): RouteFormEntry => {
+    const stops = importedGroupCounts.get(groupName) ?? 0;
+    const defaultCap = systemSettings?.default_cap;
```
ColinToft: ok so i wasn't aware of this setting and it actually looks like this is probably a mistake to have this in the settings at all (my bad). the setting I think we should be using is boxes_per_car.

frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
```diff
@@ -1,159 +1,350 @@
...
+          },
+          settings: {
+            route_start_time: routeStartDateTime(
+              row.form.routeDate,
+              row.form.startTime
+            ),
+            num_routes: row.form.routeCount,
+            return_to_warehouse: row.form.returnToWarehouse,
+            max_stops_per_route: systemSettings?.default_cap ?? null,
+            max_boxes_per_driver: systemSettings?.boxes_per_car ?? 10,
```
ColinToft: I might be wrong but isn't boxes_per_car guaranteed to exist in settings? we might be able to remove hte `?? 10`. If not let me know. I think randomly falling back to 10 without telling the user is not a great idea

frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
```diff
@@ -1,159 +1,350 @@
...
+          aria-label={`Generate routes for ${row.deliveryGroup}`}
+          className="size-4 cursor-pointer rounded-[4px] accent-blue-300"
+        />
+      ),
+    },
+    {
+      key: 'name',
+      header: 'Name',
+      render: (row) => (
+        <div className="relative w-48">
```
ColinToft, line 316: the designs mention that when you change the date, we should also update its name. can we add that in? and probably only auto-change it if the name was previously the default (like only auto-change it if it hasn't already been changed manually)

frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
```diff
@@ -1,159 +1,350 @@
...
+  checked,
+  disabled = false,
+  onChange,
+}: {
+  checked: boolean;
+  disabled?: boolean;
+  onChange: (checked: boolean) => void;
+}) {
+  return (
+    <div
```
ColinToft, line 105: For some reason this component looks a little weird right now. Do you mind taking a look at it? It's like the circle is at the right side either way.

frontend/src/pages/admin/routes/generation/ConfigureStep.tsx
ColinToft: and similar for the other fallbacks here (except for default_cap which we should remove entirely or just pass in null, we won't use this in route gen)

ColinToft (approved):
Nice work! Left a few comments but I'm good to merge once done!

PR #239, opened 2026-08-04
Implement Remember Me Functionality

kingMonkeh:
## JIRA Ticket
[Implement Remember Me Functionality in Front End](https://f4kblueprint.atlassian.net/browse/F4KRP-186)

## Implementation Description

* Implemented Remember Me functionality
* Implemented inactivity tracker to log user out if inactive for 15 mins
* Updated logout endpoint to return 204 for consistency
* Threaded remember me from backend to frontend
* Added new rememberMe cookie to keep track of state

## Steps to Test

1. Do not tick remember me when logging in
2. Be inactive for 15 minutes (if you dont want to wait you can manually update frontend/src/contexts/useInactivityTimeout.ts to change the timeout to 5 seconds and then see if you get logged out automatically
3. Try the remember me path, you shouldnt get logged out, if you close your browser and reopen you should still be able to auto login

*

## Checklist


ColinToft:
@claude review

ColinToft:
## Logout doesn't revoke tokens when the access token is expired

The logout endpoint identifies the user by decoding the `Authorization` bearer token, then calls `revoke_tokens` to invalidate all their Firebase refresh tokens. But the frontend has no refresh-on-401 interceptor — `useRefresh` runs once on mount with `staleTime: Infinity`, so the access token goes stale after ~1 hour. After that, `verify_id_token` throws `ExpiredIdTokenError`, the `except Exception` swallows it, and `revoke_tokens` never runs. Cookies are still cleared (the `finally` block handles that), so the user *appears* logged out, but their server-side session is never actually revoked.

This matters because every auth dependency in the app uses `verify_id_token(check_revoked=True)` — if revocation succeeded, a copied refresh token would be blocked on the next API call. But since revocation silently fails whenever the token is stale, the security guarantee is weaker than the code suggests.

### Suggested fix

Identify the user from the `refreshToken` cookie instead of the access token. The cookie is long-lived (not subject to the 1-hour expiry), it's already present in the request, and it's the credential being revoked — so it's the natural thing to key on. `renew_token` in `auth_service.py` already implements the pattern: exchange via `firebase_rest_client.refresh_token()`, decode `sub` from the fresh ID token, look up the user. Logout can reuse the same flow.

This also makes the endpoint more self-consistent now that it uses `optional_security` (the bearer token is truly optional).

### Separate concern: inactivity logout + revocation scope

`firebase_admin.auth.revoke_refresh_tokens` is **per-user, across all sessions** — it stamps `validSince` on the Firebase user record, which invalidates every token on every device. So if the inactivity timer fires on an idle laptop tab, it would also log the driver out on their phone mid-route (and the cross-tab issue makes this even easier to trigger).

Inactivity logout should probably just clear cookies (device-scoped, proportionate to "you walked away") and reserve revocation for the explicit Log Out button. That would mean the endpoint needs to distinguish the two cases — e.g. a query param or separate route.

ColinToft (approved):
OK this overall looks good to me! Claude has some comments that I pasted in here, because honestly I don't fully understand all the context here 😭 but I'm happy to chat more about these things.

kingMonkeh:
@claude review

kingMonkeh:
@claude review

PR #250, opened 2026-08-06
Renew an aged-out token instead of ending the session

### LINKED ISSUE #249: Add refresh-on-401 axios interceptor
## Problem

The frontend currently has no mechanism to recover from an expired access token. `useRefresh` in `frontend/src/api/auth.ts` runs once on mount with `staleTime: Infinity` and `refetchOnWindowFocus: false`, so the Firebase ID token goes stale after ~1 hour.

After that, every API call returns 401 with no recovery — the user has to manually reload the page.

## Proposed fix

Add a response interceptor in `frontend/src/lib/axiosClient.ts` that:

1. Catches 401 responses
2. Calls `/auth/refresh` to get a fresh access token
3. Updates the auth store with the new token
4. Retries the original request

Standard pattern — queue concurrent requests while the refresh is in-flight so only one refresh call happens at a time.

Currently the response interceptor in `axiosClient.ts` is a no-op pass-through, so this slots in cleanly.

## Scope

This is independent of the remember-me feature (PR #239) — it's a general session-management improvement that benefits all users.

ColinToft:
Closes #249.

## The problem

`axiosClient` treats every 401 as a dead session, on this reasoning:

> There is nothing to retry with: revocation invalidates the refresh token too, so the session is genuinely over and the only way forward is logging in again.

That is true of a **revoked** session, but a 401 is also what an **expired** token gets — and expiry is the ordinary case. Firebase mints access tokens with an hour of life, `useRefresh` runs once on mount with `staleTime: Infinity`, and nothing renews the token in between. So a tab left open over a shift outlives its token while the refresh cookie sitting beside it is still perfectly good, and the user is signed out and bounced to `/login` in the middle of whatever they were doing.

## The fix

Nothing has to guess which of the two happened, because the server already answers the question. `/auth/refresh` reads **only** the httpOnly refresh cookie — no bearer token — so asking it *is* the test: it hands back a new access token while the session is alive, and 401s once it is not.

So a 401 now refreshes and replays the request, and only a refusal from the refresh itself ends the session. A 403 is still left alone for the reason it always was.

Two details worth flagging:

- **The exchange lives in one place.** `refreshSession` is the only thing that knows how to turn the cookie into a session, and `useRefresh` now calls it too — so a page reload and a token that aged out under an open tab restore the session by the same path rather than two near-copies.
- **It is single-flight.** Each exchange rotates the refresh cookie, so a screen that discovers the stale token across six parallel requests must not spend the cookie six times; the losers of that race would be presenting a cookie the winner had already replaced.

A refresh that never gets an answer — wifi dropping in a stairwell — deliberately leaves the session standing. That is the server declining nothing; signing the driver out over it would lose whatever they were mid-way through.

## Tests

12 new cases in `axiosClient.test.ts` (36 total, all green), covering: refresh-and-replay, that the replay carries the *new* token, one refresh for a burst of six requests, session ends on a refused refresh, no second refresh when the replay is refused too, session survives a network error and a 500 from the refresh, the refresh is never itself refreshed, and no refresh for a request that carried no token.

The stub keys off the token rather than call order, so the burst case does not depend on which request wins the race.

`tsc`, `eslint --max-warnings=0`, and `prettier --check` are all clean.


ColinToft:
@claude review

ColinToft:
@claude review

kingMonkeh (approved):
Overall logic looks good, took me a while to understand but that's just cause I'm not the most familiar with Axios

Only nit I have per say is that the comments are very bloaty, I think the comments are useful but some of them can probably be made more concise

EDIT: Also seems like you'll have to fix some merge conflicts 😂

PR #251, opened 2026-08-07
F4KRP-49: Wire route generation

Kerushani:
## JIRA Ticket
[F4KRP-49](https://f4kblueprint.atlassian.net/browse/F4KRP-49)

### Route generation overview

1. **Choose a routing algorithm** -> call a single routing engine.
2. **Map engine output → DB** -> turn results into ordered `Route` / `RouteStop` rows (including polylines) and persist them under `RouteGroups`.

---

### Worker lifecycle

| Phase | Behavior |
| --- | --- |
| **Start** | Once on app lifespan via `start_route_generation_worker()` |
| **Stop** | On app shutdown |

**Loop**

1. Sleep on `_wake_event`
2. On wake (enqueue) → clear the event
3. Drain the queue: claim + run **one job at a time** until no `PENDING` jobs remain
4. Go back to sleep

**Claim**

- Atomically promote oldest `PENDING` → `RUNNING` in one `UPDATE … RETURNING`

**Dispatch**

- Resolve algorithm (`GoogleMapsFleetRoutingAlgorithm`)
- Call `run_generation_job(...)`

## Steps to Test

1. POST /jobs/generate to enqueue a job
2. The in-process worker (started on app lifespan) claims it and runs the runner
3. HTTP returns 202 immediately; the worker does the rest in-process.
4. Check if the RouteGroup is added to the table

*

## Checklist

### DIFF
--- a/backend/python/app/services/implementations/job_service.py
+++ b/backend/python/app/services/implementations/job_service.py
@@ -147,6 +147,16 @@ class JobService:
             raise error
 
     async def enqueue(self, job_id: UUID) -> None:
+        """Wake route-generation worker for a pending job.
+
+        Does not claim or run the job; that is the worker's job. A cancelled
+        (or otherwise non-pending) job is left alone so a late enqueue cannot
+        resurrect work that an admin already stopped.
+        """
+        from app.services.implementations.route_generation_worker import (
+            wake_route_generation_worker,
+        )
+
         try:
             job = await self.get_job(job_id)
 
@@ -162,13 +172,6 @@ class JobService:
                 )
                 return
 
-            job.progress = ProgressEnum.RUNNING
-            job.started_at = self.est_now_naive()
-            self.session.add(job)
-            await self.session.commit()
+            wake_route_generation_worker()
         except Exception:
             self.logger.exception("Enqueue failed for job %s", job_id)
-            try:
-                await self.update_progress(job_id, ProgressEnum.FAILED)
-            except Exception:
-                self.logger.exception("Failed to mark job %s as FAILED", job_id)
--- a/backend/python/app/services/implementations/job_service.py
+++ b/backend/python/app/services/implementations/job_service.py
@@ -11,6 +11,9 @@ from sqlmodel import col, select
 from app.models.enum import ProgressEnum
 from app.models.job import Job
 from app.schemas.route_generation import RouteGenerationGroupInput
+from app.services.implementations.route_generation_worker import (
+    wake_route_generation_worker,
+)
 
 if TYPE_CHECKING:
     from sqlalchemy.engine import CursorResult
@@ -153,10 +156,6 @@ class JobService:
         (or otherwise non-pending) job is left alone so a late enqueue cannot
         resurrect work that an admin already stopped.
         """
-        from app.services.implementations.route_generation_worker import (
-            wake_route_generation_worker,
-        )
-
         try:
             job = await self.get_job(job_id)
 
--- a/backend/python/app/services/implementations/job_service.py
+++ b/backend/python/app/services/implementations/job_service.py
@@ -174,3 +174,4 @@ class JobService:
             wake_route_generation_worker()
         except Exception:
             self.logger.exception("Enqueue failed for job %s", job_id)
+            raise
--- a/backend/python/app/services/implementations/route_generation_runner.py
+++ b/backend/python/app/services/implementations/route_generation_runner.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import time
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, cast
 from zoneinfo import ZoneInfo
@@ -79,6 +80,7 @@ async def run_generation_job(
     cancelled along the way, in which case the cancellation stands and
     nothing is saved.
     """
+    started = time.perf_counter()
     try:
         job = await _load_running_job(session, job_id)
         if job is None:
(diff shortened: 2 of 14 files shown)


Kerushani:
@claude review

backend/python/app/services/implementations/job_service.py
```diff
@@ -147,6 +147,16 @@ async def cancel_job(self, job_id: UUID) -> Job | None:
             raise error
 
     async def enqueue(self, job_id: UUID) -> None:
+        """Wake route-generation worker for a pending job.
+
+        Does not claim or run the job; that is the worker's job. A cancelled
+        (or otherwise non-pending) job is left alone so a late enqueue cannot
+        resurrect work that an admin already stopped.
+        """
+        from app.services.implementations.route_generation_worker import (
```
ColinToft: Is this intentionally imported down here? I'm wondering if we can just import at the top

ColinToft: but if this is intentional like if it has lazy initialization or something that's fine!

backend/python/app/services/implementations/route_generation_runner.py
```diff
@@ -0,0 +1,435 @@
...
+    algorithm: RoutingAlgorithmProtocol,
+) -> None:
+    """Generate and save the routes one job asked for.
+
+    Ends with the job either `COMPLETED` (with a `RouteGroup` linked and the
+    summary recorded) or `FAILED` (with `error_message` set), unless it was
+    cancelled along the way, in which case the cancellation stands and
+    nothing is saved.
+
+    TODO(cancel): Honor POST /jobs/{id}/cancel cooperatively via the jobs row:
```
ColinToft: What are these TODO (cancel) things? They kinda look to me like things that you have already implemented?

backend/python/app/services/implementations/route_generation_worker.py
```diff
@@ -0,0 +1,199 @@
...
+        return None
+
+    await session.commit()
+    return job_id
+
+
+async def recover_route_generation_jobs(session: AsyncSession) -> None:
+    """Fail jobs left RUNNING across a restart, and wake if PENDING remain.
+
+    A RUNNING job cannot be resumed: the Google call is gone with the old
```
ColinToft, line 94: It might be nice if we were able to resume it. Is there any particular information that would be lost that makes it unresumable? Also down to chat about this if there is, we can see about maybe including the info in the job object or similar. Although that said, maybe not worth spending thatttt much time on this because hopefully we won't get that many crashes :)

backend/python/app/services/implementations/route_generation_runner.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+class GenerationFailed(Exception):
+    """A job that can't go on, carrying a reason worth showing an admin.
+
+    Caught by `run_generation_job`, which stores the message on the job.
+    """
+
+
+def _now_est_naive() -> datetime:
```
ColinToft: We should consider making this into a shared util function somewhere, I feel like this already exists in the codebase maybe? And also twice in this PR

backend/python/app/services/implementations/route_generation_runner.py
```diff
@@ -0,0 +1,435 @@
...
+
+    from sqlalchemy.engine import CursorResult
+    from sqlalchemy.ext.asyncio import AsyncSession
+
+    from app.schemas.route_generation import RouteGenerationSettings
+    from app.services.protocols.routing_algorithm import RoutingAlgorithmProtocol
+
+logger = logging.getLogger(__name__)
+
+GENERATION_TIMEOUT_SECONDS = 180.0
```
ColinToft, line 55: Where did this number come from? Did you test it on the set of ~500 routes?

backend/python/app/services/implementations/route_generation_runner.py
```diff
@@ -0,0 +1,435 @@
...
+        if location.latitude is not None and location.longitude is not None
+    ]
+
+    if not routable:
+        raise GenerationFailed(
+            f"Location group '{group.name}' has no geocoded locations to route."
+        )
+
+    if len(routable) < len(locations):
+        logger.warning(
```
ColinToft: This should not just be a logged warning - ideally we would geocode them here if they aren't already, but at least raise an error. I don't thinkkkkk this should ever happen? So it's maybe ok to just error, but definitely we want something to happen here

ColinToft (changes requested):
Thanks for making this PR, and overall things are looking great! I left a couple comments for some things I'd want to change before merging. Good work on this!

backend/python/app/services/implementations/route_generation_runner.py
Kerushani: Let me get back to you on this. - I did not test this on ~500 routes - but is this the number it should be able to handle?

`google_maps_routing_service.py` has an individual API timeout of 60 seconds and the runner here also has a cap (`ENGINE_TIMEOUT_SECONDS`) of 120 seconds.

GENERATION_TIMEOUT_SECONDS should account for the whole generation step - that being the async steps and building the route group.

So roughly I was looking at 180 = 120 routing engine step + ~60s slack for `_build_route_group` (including polyline fetches which also uses the Google Maps API).

backend/python/app/services/implementations/route_generation_runner.py
Kerushani: Yup I agree, let me make that an error

backend/python/app/services/implementations/job_service.py
Kerushani: No, not required to be imported down here and should be moved to the top

backend/python/app/services/implementations/route_generation_worker.py
Kerushani: That's a good point. The job should have the input payload since when its picked up, its not deleted. 

An option to go about resuming: 

Auto-retry: On startup, flip orphaned RUNNING → PENDING again with idempotency (“don’t create a second RouteGroup for the same job” because right before a crash the RouteGroup might have been completed. 

So right before marking the job as PENDING, I'd check if it has a route_group_id set. If there is, I would mark it as completed and if not, I'll remark it as PENDING and retry once.

Kerushani: Have the changes for it - but let me know if i can make any changes to it -> https://github.com/uwblueprint/food4kids/pull/255

backend/python/app/services/implementations/route_generation_runner.py
Kerushani: Agreed; pulled this into app/utilities/datetime_utils.py and switched the runner/worker over.

backend/python/app/services/implementations/route_generation_runner.py
Kerushani: ok actually 3 mins is too long

backend/python/app/services/implementations/route_generation_runner.py
Kerushani: Yeah, these are already implemented (Running-only save/fail updates + skip if no longer Running). And `cancel_job()` seems to handle it as well. I’ll remove the stale TODO(cancel) comments.

ColinToft (approved):
Nice, thanks for fixing all these!!

PR #276, opened 2026-08-18
F4KRP-35: Admin Driver Management Page

aaronwxng:
## Summary
- build the Figma-aligned driver management table, add-driver invite modal, and driver side panel
- add server-side name search, sorting, pagination, derived activity, mileage, and last-delivery aggregates
- support profile editing, driver notes, mail links, deletion confirmation, current-week assignments, and completed-route history
- make invite-only profile fields optional and migrate availability from seven days to Monday-Friday
- regenerate the OpenAPI client; no Postman files changed

## Validation
- backend full suite: 832 passed
- frontend tests: 64 passed
- frontend lint and production build passed
- backend Ruff passed
- Alembic reports a single head and the migration applied successfully

## Post-rebase note
After rebasing onto latest main, frontend lint/build and backend Ruff passed again. Focused pytest could not boot in the stale local backend image because latest main added google.cloud.bigquery; the image rebuild stalled fetching Docker Hub metadata. GitHub CI will validate with a fresh dependency install.


frontend/src/pages/admin/drivers/components/DriverPanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+}
+
+function currentWeek() {
+  const now = new Date();
+  const monday = new Date(now);
+  const day = now.getDay() || 7;
+  monday.setDate(now.getDate() - day + 1);
+  const sunday = new Date(monday);
+  sunday.setDate(monday.getDate() + 6);
+  return { start: isoDate(monday), end: isoDate(sunday) };
```
eddywang4340, line 34: I think it would be better to use `toNaiveDateString()` instead of defining `isoDate` here. This function should be used in this repo

PR #284, opened 2026-08-24
F4KRP-194: Add driver/admin annoucement email flags

Kerushani:
## JIRA Ticket
[F4KRP-194](https://f4kblueprint.atlassian.net/browse/F4KRP-194)

These are backend changes to systems settings. This does not implement the settings page frontend. I'll add that in a separate PR.

## Implementation Description

* Adds `announcement_emails_to_admins` and `announcement_emails_to_drivers` to system settings.
* Modifies existing `send_announcement_emails_to_drivers` to `send_announcement_emails` so that it checks the admin's `announcement_emails_...` flags when sending an email
* If `announcement_emails_to_admins` is enabled email is just sent to admins, if `announcement_emails_to_drivers` is enabled email is just sent to driver ; if both are enabled, everyone gets an email ; if flags are false, then no email is sent to everyone

Note on recipients: the announcement's author is excluded from the send. Previously this happened implicitly, only drivers were emailed and only admins can trigger a send, so an author was never a recipient. Now that admins receive announcement emails, admins who write the announcement are skipped.

Also adds twitter column in migration and associated files.

## Steps to Test
1.

*

## Checklist


Kerushani:
> Duplicate emails if a user is ever both an admin and a driver

I can deduplicate by user_id in case there's changes in the future.

> Silent behavior change: the announcement author is now excluded from recipients.

Makes sense that the author wouldn't get the email - I would say that is the intended behaviour.

the code block:
```
for user, role in recipients:
    if user.user_id == announcement.user_id:
        continue
```
prevents announcement authors from getting the email.

> `Admin.receive_email_notifications` is still unused.

Unsure if this has future use so i'll leave it alone for now

PR #292, opened 2026-08-29
Unified TimePicker: one component, clock everywhere, padding as a prop

ColinToft:
Makes `TimePicker` match the design and serve both call sites.

- One scrollable list of times on a half-hour step, replacing the three hour/minute/AM-PM wheels.
- Clock icon everywhere, following the newer route-generation frames.
- `padding` prop, 12 (route generation, default) or 24 (Settings), shared by trigger and options so their text aligns.
- Trigger accepts typing — how staggered times like 8:45 are entered. Off-step values merge into the list.
- Opening centres the selection.
- Panels stay pinned to the trigger while scrolling: layouts provide a `PopupHost` in the scrolled content that popovers portal into, so they scroll with the page (also fixes `DatePicker`).

Judgment calls, easy to overrule:

- **A bare hour falls in the delivery day**, 8 AM–7:59 PM: `1` is 1 PM. Explicit AM/PM wins.
- **Bad input** reddens while typing, reverts on blur; cleared reverts unmarked.
- Trigger radius `rounded-xl`, per design.


ColinToft:
## Follow-up: the panel now takes the padding too

Colin: *"popover should also have the same padding — the goal is for the options in the dropdown to perfectly align with the typed input field in the trigger above."*

I'd originally applied the padding to the trigger only. That was wrong; fixed in 1f7c7e26. I measured this in Chrome rather than reasoning about it, and two things the arithmetic missed turned out to matter.

**No 1px correction is needed.** The trigger's ring is an `outline` with `outline-offset-[-1px]`, and outlines take no layout space, so the trigger's text sits at exactly `padding` from its border box. `PopoverContent` has no border, and `align="start"` puts its left edge on that same border box. So the two coordinate systems already share an origin.

**The real blocker was the option pill, not the padding.** Options were centred in a `min-w-12` (48px) pill, and every label is narrower than that, so centring offsets the label by half the leftover width — a different offset for `9` than for `12`. No single panel padding could ever have lined them up. The pill now hugs its label (`px-3`, no `min-w`), which fixes the alignment *and* keeps the pill symmetric about its own text; left-aligning inside the fixed 48px pill, which I tried first, put a short label visibly off-centre in its circle.

Panel padding is therefore `padding - 12`, with the pill's own `px-3` supplying the remainder.

**Measured, both variants, after the entry animation settles:**

| padding | trigger text x | option text x | delta |
| --- | --- | --- | --- |
| 12 | 540.67 | 540.50 | -0.17px |
| 24 | 805.33 | 805.50 | +0.17px |

Every option in the first column lands on the same x (12 hours, one value), so it holds for one- and two-digit labels alike. The 0.17px residue is floating-ui rounding the panel to a device pixel against a trigger that the style guide's centred layout leaves on a fractional x; in the route-generation table the trigger is on an integer x.

### One artifact worth a look

At `padding=12` the panel supplies zero padding, so the selected chip is **flush against the panel's left edge** — visible on the left picker in the screenshot. This is forced by the spec rather than by the implementation: the option's text has to be 12px from the panel's edge, and the pill needs ~12px of its own padding to look round, which leaves nothing between the pill and the edge. Giving the panel breathing room means either shrinking the pill's padding (the chip stops looking round) or letting the panel hang left of the trigger (the boxes themselves stop lining up). At `padding=24` there's 12px of room and it looks fine. Happy to trade a pixel of alignment for a little air here if you'd rather.

### Also

`open` replaces the internal-only open state, so the style guide can pin both variants open at once — Radix closes a non-modal popover as soon as focus lands in another, so two self-opening pickers wouldn't both stay up. The style guide's Padding preview now shows both panels open, since the panel is the thing being demonstrated.

Panel padding is tested through the pure module that decides it (`TimePicker.padding.ts`) — the panel renders through a Radix portal and so is absent from `renderToStaticMarkup` — with an invariant test (`panel + 12 === trigger`) rather than just asserting the literals, so moving one side without the other fails.

Unrelated, noticed while testing and **not** fixed here: opening the picker doesn't scroll the columns to the selected value, so at 9:30 you open onto `12, 1, 2…` with the selected `9` out of view. Pre-existing; say the word and I'll pick it up separately.

ColinToft:
## Rebuilt as a single list of whole times

Colin: *"yours has separate hour/minute sliders, the slack essentially models it as a single dropdown with times spaced every half hour."* Correct — the three-column wheel was a different component. Rebuilt in 0567088d.

I read both frames over the Figma REST API rather than guessing:

| | trigger `paddingLeft` | menu row `paddingLeft` | options |
| --- | --- | --- | --- |
| Route generation → Start Time | 12 | 12 | 9:00 / 9:30 / 10:00 AM |
| Settings → Route Reminders → "At what time?" | 24 | 24 | 8:30 / 9:00 / 9:30 AM |

(Settings' raw numbers are 25.27 — that instance is scaled 1.053; `cornerRadius` 16.85 and `paddingTop` 12.64 divide out the same way.)

**Both call sites want the same option set,** so the half-hour step stays internal rather than becoming a prop. Rows are 12px padding around 18px text = 42px and the menu is 126px, i.e. the three visible rows both designs show — that's what `py-3` + `max-h-32` reproduce.

**The trigger and the options now take the same padding class,** so alignment falls out of the structure instead of being arranged. That also retires the flush-chip trade-off I flagged last round: it only existed because a pill needed its own padding inside a fixed-width cell, and a full-width row has no such constraint. `timePickerPaddingClasses` collapsed back to a single `timePickerPaddingClass`.

**On the staggered times you asked me to check first** — the designs do show `8:45 AM`, `10:05 AM`, `9:45 AM` in the Start Time column, in the same column whose open menu offers half hours. So they're stored values, not options. A value that doesn't fall on the step is merged into the list in order and stays visible as the selection; `timeOptions('08:45')` returns 49 entries with `08:45` at index 18. That covers the design without inventing a second option set.

**Bonus finding:** the newer route-generation frames carry a `clock` instance where the older ones have `arrow_drop_down`. Gurman's clock mock *did* land — the icon now matches the design, not just Isabelle's stated intent. My earlier note on the PR was out of date.

### Scroll to selection

Opening the picker centres the selection, clamped at both ends. Verified end-to-end in Chrome, not just in the unit tests:

| | scrollTop | |
| --- | --- | --- |
| 9:30 AM (mid-list) | 755 | centred, 43px from the top of a 128px box = (128−42)/2 |
| 12:00 AM (first) | 0 | flush top |
| 11:30 PM (last) | 1888 | = the maximum, flush bottom |

This caught a real bug. The scroll box was reached through a `ref` on the list, but React assigns a child's ref **before** its parent's, so `listRef.current` was still `null` when the selected option's callback ref fired — the list silently never scrolled (`scrollTop: 0`, selection off-screen). It now reads the list off the option itself, which has no ordering dependency.

### Alignment, re-measured on the new structure

| padding | trigger text x | option text x | delta |
| --- | --- | --- | --- |
| 12 | — | — | −0.17px |
| 24 | — | — | +0.17px |

Panel width now tracks the trigger (`--radix-popover-trigger-width`), and computed padding is 12px/12px and 24px/24px on trigger and option respectively. The 0.17px residue is floating-ui rounding the panel to a device pixel.

### One thing I changed that you didn't ask for

The trigger's radius went `rounded-sm` → `rounded-xl`. The design specifies r=16 on the Start Time `Text Box`, and `Dropdown.tsx` already uses `rounded-xl` for r16 by the same convention; a `rounded-sm` trigger above a `rounded-xl` panel looked inconsistent once they sat flush. Easy to revert if you'd rather keep it separate.

### Still open

One route-generation row in the design shows a **text caret** after `9:00 AM`, and the node is named `Text Box` — so the trigger may be meant to accept typing, which would be how 8:45/10:05 get entered in the first place. I did **not** build free text entry; that's a real scope call (parsing, validation, what happens on a bad string) rather than something to slip in. Say the word and it's its own PR.

Also unchanged: `max-h-32` gives exactly the 3 rows the route-generation design shows (126px). Settings' rows are taller in Figma (48px), so that variant shows ~2.7 rows and clips the third — a deliberate scroll affordance, but one class if you want it different.

CI green on all four jobs.

PR #293, opened 2026-08-29
Keep drivers and admins out of each other's app

ColinToft:
Drivers and admins could walk into each other's apps: `/admin` was mounted with no guard, and `AuthProvider` only checked `isAuthenticated`, never `user.role`. Two backend endpoints were open to any signed-in driver for the same reason — `GET /drivers/` (every volunteer's phone, home address, plate, email) and `POST /jobs/generate`. Both are now `ADMIN_ONLY`; verified nothing driver-side calls either.

A new `RequireRole` wraps `/admin/*` and `/driver/*`. Judgment calls: a user in the wrong app is silently redirected to their own home rather than shown a 403 (`ForbiddenPage` stays dev-only), and the decision fails closed — `role` is a free-form string on the backend, so anything that isn't `admin` or `driver` is admitted to neither app. Review follow-ups made the whole `/jobs` and `/location-groups` routers admin-only too — both are only used by the admin generation wizard.

Component tests run on happy-dom, the repo's one test environment.


backend/python/app/routers/driver_routes.py
```diff
@@ -49,10 +48,14 @@ async def get_drivers(
     session: AsyncSession = Depends(get_session),
     driver_id: UUID | None = Query(None, description="Filter by driver ID"),
     email: str | None = Query(None, description="Filter by email"),
-    _auth: bool = Depends(require_driver_or_admin),
+    _auth: bool = Depends(require_admin),
```
eddywang4340, line 51: More of a question but I see that the function `complete_driver_registration` doesn't have an `_auth`. I feel like completing a driver registration should be admin-only operation so should we add a `Depends(require_admin)` there as well?

backend/python/app/routers/job_routes.py
```diff
@@ -36,9 +36,12 @@ async def get_jobs(
 async def generate_job(
     req: RouteGenerationGroupInput,
     service: JobService = Depends(get_job_service),
-    _auth: bool = Depends(require_driver_or_admin),
+    _auth: bool = Depends(require_admin),
```
eddywang4340, line 39: Same thing for `get_jobs`. It is `Depends(require_driver_or_admin)` and I feel like it should be admin-only (unless I'm missing something)

eddywang4340:
Also for `location_group_routes` all endpoints right now are `Depends(require_driver_or_admin)` meaning that drivers are able to create, update, and delete location groups. Probably should make it `require_admin`

eddywang4340 (approved):
Added some comments above about the other endpoint routes! But everything else LGTM!

backend/python/app/routers/job_routes.py
ColinToft: Agreed — made all of `/jobs` admin-only in e881edad (incl. `GET /jobs/{job_id}`, which the generation wizard polls).

backend/python/app/routers/driver_routes.py
ColinToft: This one needs to stay public: it is the create-password step from the invite email, before the driver has a Firebase account, so there is no bearer token to check yet. The unused/unexpired invite ID (row-locked) is what authorizes it.

ColinToft:
Good catch on `/location-groups` — whole router is admin-only now (26624c9d).

PR #296, opened 2026-08-29
One source of truth for per-car capacity, and no silent defaults

ColinToft:
`RouteGenerationSettings.max_boxes_per_driver` defaulted to 14 while `SystemSettings.boxes_per_car` defaults to 10, and the configure screen posted the setting with no loading guard — so if the query hadn't resolved, the key fell out of the JSON and the schema default of 14 silently replaced the configured 10. Same shape for `service_time_minutes` and `children_per_box`.

Defaults now live only in `app/models/system_settings.py`. The three schema fields are required, so omitting one is a 422 rather than a substitution; `cluster_locations` takes the cap as a required argument. `ConfigureStep` waits for settings and builds the numbers through `routeCapacity()`, which returns `null` rather than a partial object. `sweep_algorithm.SweepAlgorithm` had zero importers and is deleted.

Question for you: is real per-car capacity 10? Schema said 14, settings default says 10. This standardizes on 10 — if drivers fit 14, just set it in Settings.


Kerushani:
> Question for you: is real per-car capacity 10? Schema said 14, settings default says 10. This standardizes on 10 — if drivers fit 14, just set it in Settings.

Valid question - I had `RouteGenerationSettings.max_boxes_per_driver` defaulted to 14 because of some changes requested by Emily. -> https://f4kblueprint.atlassian.net/browse/F4KRP-177

Kerushani (approved):

PR #297, opened 2026-08-29
Run CI on Node 24.13.0 to match the Dockerfile

ColinToft:
CI was pinned to Node 20.11.1 while `frontend/Dockerfile` runs node:24.13.0 — prod, local dev and every contributor's machine are on 24, and only the gate was on a February 2024 build. That bit us today: jsdom 28 pulls `html-encoding-sniffer@6`, which `require()`s an ESM-only package, and `require(esm)` only landed in Node 20.19. Every local check passed and CI went red.

`node-version` becomes `24.13.0` in the three places it appears (`lint.yml`, and both jobs in `openapi-sync.yml`). Those are the only Node pins in the repo. The stale "Node 20+" note in `vite.config.ts` is rewritten.

Worth knowing: this PR's own CI does **not** exercise the `openapi-sync` bump — that job's path filter watches files this PR doesn't touch, so it reports green without running Node. The generator determinism was checked locally instead; the daily audit job covers it on main.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -43,7 +43,7 @@ jobs:
         if: steps.changes.outputs.frontend == 'true'
         uses: actions/setup-node@v4
         with:
-          node-version: "20.11.1"
+          node-version: "24.13.0"
           cache: "pnpm"
           cache-dependency-path: frontend/pnpm-lock.yaml
 
--- a/.github/workflows/openapi-sync.yml
+++ b/.github/workflows/openapi-sync.yml
@@ -95,7 +95,7 @@ jobs:
         if: steps.changes.outputs.contract == 'true'
         uses: actions/setup-node@v4
         with:
-          node-version: "20.11.1"
+          node-version: "24.13.0"
           cache: "pnpm"
           cache-dependency-path: frontend/pnpm-lock.yaml
 
@@ -206,7 +206,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "20.11.1"
+          node-version: "24.13.0"
           cache: "pnpm"
           cache-dependency-path: frontend/pnpm-lock.yaml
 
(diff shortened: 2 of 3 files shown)


ColinToft:
## Verification results

**Full frontend pipeline in a `node:24.13.0` container** (pnpm 9, matching `pnpm/action-setup@v4`
`version: 9`), run against this branch before pushing — exit 0:

| step | result |
|---|---|
| `pnpm install --frozen-lockfile` | 986 packages, lockfile up to date, no resolution changes |
| `pnpm lint` | clean, `--max-warnings=0` |
| `pnpm format:check` | "All matched files use Prettier code style!" |
| `pnpm test` | 4 files, **85 tests passed** |
| `pnpm build` | `tsc -b` clean, 4413 modules, `✓ built` |

No pre-existing failure was surfaced by the newer Node — the suite is as green on 24.13.0 as it
was on 20.11.1.

**OpenAPI generator determinism.** Since this PR's CI can't exercise the `openapi-sync` bump
(see below), both Node-dependent stages were run under 24.13.0 and diffed against what's
committed:

- `scripts/fetch-openapi.mjs` → `frontend/openapi.json` — **byte-identical**
- `openapi-ts` (incl. its Prettier post-process) → `frontend/src/api/generated` — **byte-identical**

So the bump can't introduce contract drift.

## What CI here actually proves

Being precise, because three of the four checks are green for reasons unrelated to Node:

| check | result | what it means |
|---|---|---|
| `run-lint` | **pass, 54s** | The real signal. Log confirms `node: v24.13.0`; lint, format, test and build all ran. |
| `test` (pytest) | pass, 34s | Gated on `python-backend` paths — steps skipped, no Node involved. |
| `boot` | pass, 22s | Same `python-backend` gate — steps skipped. |
| `openapi-sync` | pass, 9s | `contract` path filter doesn't match this PR, so its steps skipped. Covered by the local determinism check above. |
| `openapi-drift-audit` | skipping | Schedule-only; runs filter-free on main after merge. |

One incidental data point: the CI build finished in **8.3s** where the same build took 5m52s in
my local container — that's macOS bind-mount I/O, not Node, and worth not misreading as a
regression.

PR #299, opened 2026-08-29
Admin account bootstrapping: a non-destructive CLI + role-agnostic registration

ColinToft:
> **Stacked on #276, so CI will not run on pushes** — all four workflows are gated on `main`. Verified locally (ruff, mypy, full pytest).

The only code building an `Admin` row lived inside `seed_database()`, which deletes every table first. Now a dev runs `python -m app.create_admin`; it emails the staff member the same `account-creation` invite drivers get, and they set their own password. The send happens inside the insert transaction, so a failed email rolls the account back — no printed-link fallback. The placeholder is renamed `Driver_Name_To_Replace` → `Name_To_Replace` since it now addresses admins too.

Design call worth reviewing: rather than an admin sibling of `POST /drivers/register`, this is one role-agnostic `POST /auth/register` with `/drivers/register` deleted — the driver endpoint was already correct for an admin except its final `DriverRead` line.

Also: `admin_phone` is now optional (migration `e7b21f4a9c53`). An invite link is an admin account for 48 hours.


ColinToft:
### Note for whoever merges `main` into this branch next

The `ROUTE_POLICIES` hunk in `backend/python/tests/test_auth_integration.py` conflicts, and **the union is the wrong resolution** — both sides are *deletions*, not additions.

The merge base carried both entries:

```python
("POST", "/drivers/register"): Policy.PUBLIC,
("POST", "/drivers/test-event-email"): Policy.PUBLIC,
```

- **This branch** removed `/drivers/register` — that's the point of the PR; driver-specific registration is replaced by the role-agnostic `POST /auth/register`, which the table already carries.
- **`main`** removed `/drivers/test-event-email` in #301 (dev-scratch surface removal).

So the correct resolution drops *both*, keeping this branch's rewritten comment above `/drivers/initialize`. Taking the union re-adds two entries for routes that no longer exist on either side.

Verified against the merged app code: neither path is registered on the router, and a fresh schema dump has 48 paths including `/auth/register` and neither of those two. Full suite: 904 passed.

(Resolved this way in `d420361b`.)

ChloeW125 (approved):
LGTM!

PR #304, opened 2026-08-30
Show the route group on driver-facing views

ColinToft:
Generated route names restart at `Route 1` in every route group, so on the driver's views — where the name is the only label — it doesn't say which route it is. The group already has a meaningful name, so this shows it on the two driver-facing views that lacked it: the home card subtitle, and the route detail page (and therefore the printed PDF). Nothing is renamed or written.

Putting it into the stored name was worse: it duplicates the date at exactly the sites it was meant to help, it clips inside `UnassignedRoutePreviewCard`, and it writes the decision to the database with no backfill.

One tradeoff worth a look: the home card's formatted `Sep 25` is now replaced by `group_name`, which carries the date only by convention. Rename a group to something dateless and the home card loses its date. Keeping both prints it twice.

### DIFF
--- a/backend/python/app/models/route.py
+++ b/backend/python/app/models/route.py
@@ -133,13 +133,16 @@ class RouteDetailRead(RouteRead):
     Stops are assembled with snapshot-over-live precedence. See
     RouteStopDetailRead.
 
-    drive_date is sourced from the route's RouteGroup (mirrors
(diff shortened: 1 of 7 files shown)


ColinToft:
Closing per Colin: drivers care less about which route is which than admins do, so the driver-facing views don't need the group name. The admin-side ask this surfaced — sorting the route column numerically instead of as text — is being handled separately.

PR #306, opened 2026-08-30
Render nothing during the session restore, and stop network blips signing people out

ColinToft:
**The "Restoring your session…" flash.** The access token is memory-only, so every load of a non-public route blocks on `POST /auth/refresh` — and `/` isn't public, so even a logged-out first visit saw the gate on its way to `/login`. `AuthProvider` now renders `null` for a 300 ms grace window and explains itself only after it.

**A network blip logged you out.** `useRefresh` called `clearAuth()` on *any* refresh failure, signing out someone whose refresh cookie was still good. Now only a refusal does: any 4xx from `/auth/refresh`, which reads nothing but that cookie. A 5xx or a dropped connection leaves the session standing and offers "Try again". The interceptor keeps its 401-only rule — a 403 there means "not yours".

The refresh carries a timeout, so a hung connection reaches the retry. Not global: an import geocodes every new address inside one request.

Component tests run on happy-dom.


frontend/src/lib/axiosClient.ts
```diff
@@ -31,6 +35,20 @@ axiosClient.interceptors.request.use((config) => {
...
 
+/**
+ * Any 4xx from `/auth/refresh` means the server looked at the refresh cookie
+ * and rejected it, so retrying cannot help. A 5xx or a network error says
+ * nothing about the cookie.
+ */
+export function isRefreshRefusal(error: unknown): error is AxiosError {
+  const status = isAxiosError(error) ? error.response?.status : undefined;
+  return status !== undefined && status >= 400 && status < 500;
+}
```
kingMonkeh, line 46: nit: can we do something like

```
export function isRefreshRefusal(error: unknown): error is AxiosError {
  if (!isAxiosError(error)) return false;
  const status = error.response?.status;
  return status !== undefined && status >= 400 && status < 500;
}
```

i feel like its easier to read this way

frontend/src/contexts/AuthProvider.tsx
```diff
@@ -43,13 +57,30 @@ export function AuthProvider({ children }: { children: React.ReactNode }) {
...
+            Try again
+          </Button>
         </div>
+      );
+    }
+
+    return (
+      <div className="flex h-screen w-screen items-center justify-center bg-gray-50">
+        <span className="text-sm font-medium text-gray-500">
+          Restoring your session...
```
kingMonkeh, line 82: unrelated, but should we ask design for a loading screen for this state? or is this more of an extra thing we can think about later?

kingMonkeh (approved):
Overall looks good 😎  left 2 comments, one is a nit, the other is more of a design question

